PR #3, opened 2020-01-04
Show all/successful projects

jpopelka:
This is just a proof of concept - to show that we have such data.

### DIFF
--- a/main.py
+++ b/main.py
@@ -22,8 +22,7 @@ def node_modules(filename):
 @app.route("/")
 def main():
     return render_template(
-        "main_frame.html",
-        content=render_template("information.html"),
+        "main_frame.html", content=render_template("information.html"),
     )
 
 
@@ -44,17 +43,6 @@ def return_json(url, method="GET", **kwargs):
     return output
 
 
-def return_json_all_pages(url, limit=10, method="GET", **kwargs):
-    output = []
-    for counter in range(1, limit):
-        url += f"?page={counter}&per_page=10"
-        tmp_output = return_json(url, method, **kwargs)
-        if not tmp_output:
-            break
-        output += tmp_output
-    return output
-
-
 @app.route("/status/")
 def status():
     content = render_template(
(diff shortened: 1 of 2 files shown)


jpopelka:
Snímek z 2020-01-04 18-53-55

templates/projects.html
```diff
@@ -1,3 +1,7 @@
-<h2> Total project Counter {{ counter }} </h2>
+<h2> Total projects {{ counter }} </h2>
 
-{{ projects|safe }}
\ No newline at end of file
+{{ all_projects|safe }}
```
lachmanfrantisek: What about this:

```suggestion
{% for repo in all_projects %}
<div class="pf-c-card">
    <div class="pf-c-card__header pf-c-title pf-m-md">{{repo}}</div>
    <div class="pf-c-card__body">https://project.url and other info</div>
</div>
{% endfor %}
```

You need to remove `str` conversion in `render_template` call.

lachmanfrantisek: btw, we have some probably some "{repository_name}" value in redis:

Screenshot from 2020-01-05 12-04-50

main.py
```diff
@@ -44,17 +44,6 @@ def return_json(url, method="GET", **kwargs):
     return output
 
 
-def return_json_all_pages(url, limit=10, method="GET", **kwargs):
```
jscotka: why this method is removed? I think that the method that you wrote ``all_builds`` could use some generic ones, like this, and just change URL, becasuse lots of pages there are paginated, so it needs to read all pages for complete information.

jpopelka: fixed - there's more generic `all_from()` function now

jpopelka:
@lachmanfrantisek nice suggestions, however, this was really meant as a proof of concept to show just the **number** of projects rather than some fancy table. It certainly needs a complete redesign, but that's up to some other PR/work. 

> I vote to merge anything useful immediately

+1, merging now

PR #4, opened 2020-01-06
Define what Total/Successful project means

jpopelka:


TomasTomecek:
That's a pretty long header: how about set it as a description, or just a line below the header.

jpopelka:
added as a line below the header

PR #6, opened 2020-02-06
deployment scripts created

sakalosj:


Makefile
```diff
@@ -1,6 +1,22 @@
...
 	FLASK_ENV=development python3 main.py
+
+run_docker:
+	docker run -p 443:8443 -p 80:8080 -v $(DIR)/secrets:/secrets -i usercont/packit-dashboard:stg
+
+build_base:
+	docker build -t usercont/packit-dashboard-base -f Dockerfile.base .
+
+build:
+	docker build -t usercont/packit-dashboard:stg -f Dockerfile .
```
jpopelka: add `--rm` please to remove temporary container(s)

sakalosj: Main difference is that if code is updated, you will not  go thorough package installation process every time - will speed up deployment, local  testing. It was especially useful during fine tuning of docker image. Also i think is simpler, than using another configuration layer.

sakalosj: --rm added

jpopelka:
> Main difference is that if code is updated, you will not go thorough package installation process every time - will speed up deployment, local testing. It was especially useful during fine tuning of docker image. Also i think is simpler, than using another configuration layer.

Ok, I see your point, however, `docker build` uses cache so if a Dockerfile instruction hasn't changed since last build, the layer is taken from cache, see all the ` ---> Using cache` in 'docker build' output.
So for local "fine tuning" there's no difference in build time and when we build in a hub/repository we always want to install requirements from scratch to pull in updated packages. And I don't think it's simpler to juggle with two images instead of one.

* https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#leverage-build-cache
* https://thenewstack.io/understanding-the-docker-cache-for-faster-builds/

sakalosj:
> > Main difference is that if code is updated, you will not go thorough package installation process every time - will speed up deployment, local testing. It was especially useful during fine tuning of docker image. Also i think is simpler, than using another configuration layer.
> 
> Ok, I see your point, however, `docker build` uses cache so if a Dockerfile instruction hasn't changed since last build, the layer is taken from cache, see all the ` ---> Using cache` in 'docker build' output.
> So for local "fine tuning" there's no difference in build time and when we build in a hub/repository we always want to install requirements from scratch to pull in updated packages. And I don't think it's simpler to juggle with two images instead of one.
> 
> * https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#leverage-build-cache
> * https://thenewstack.io/understanding-the-docker-cache-for-faster-builds/

you are right, i am aware o layers, I checked old Dockerfile and it was not configured properly so it was installing packages all the time - i thought that dnf will be excuted evry time to ensure latest version are included every time. By simpler i meant to not use ansible. 

Thanks for pointing out, this is ideal to have one docker file.

jpopelka:
> By simpler i meant to not use ansible.

My "best practice" is to do only the necessary minimum in a `Dockerfile`, i.e. copy all the scripts/file/playbooks in and run them and keep all the provisioning logic in those scripts/playbooks.

So if you want to be consistent with the rest of our repos you should put the `dnf install` in one playbook (see [this](https://github.com/packit-service/packit-service/blob/master/files/install-deps.yaml)) and the `mkdir ... && chmod a+rwx ...` (see [this](https://github.com/packit-service/packit-service/blob/master/files/tasks/common.yaml) and [this](https://github.com/packit-service/packit-service/blob/master/files/tasks/httpd.yaml)) in another.

But I think I can accept it as it is if you remove the make `build_base` target.

sakalosj:
> > By simpler i meant to not use ansible.
> 
> My "best practice" is to do only the necessary minimum in a `Dockerfile`, i.e. copy all the scripts/file/playbooks in and run them and keep all the provisioning logic in those scripts/playbooks.
> 
> So if you want to be consistent with the rest of our repos you should put the `dnf install` in one playbook (see [this](https://github.com/packit-service/packit-service/blob/master/files/install-deps.yaml)) and the `mkdir ... && chmod a+rwx ...` in another (see [this](https://github.com/packit-service/packit-service/blob/master/files/tasks/common.yaml) and [this](https://github.com/packit-service/packit-service/blob/master/files/tasks/httpd.yaml)).
> 
> But I think I can accept it as it is if you remove the make `build_base` target.

i have reworked it to follow best practices, please check

files/cfg/httpd-packit.conf
```diff
@@ -0,0 +1,27 @@
...
+  DocumentRoot /usr/share/packit_dashboard/
+  ServerName dashboard.stg.packit.dev
+  ServerAdmin user-cont-team@redhat.com
+
+  <Directory /usr/share/packit_dashboard>
+      Require all granted
+  </Directory>
+
+
+  WSGIDaemonProcess packit threads=5
```
jpopelka: Maybe we don't need that many threads. I'd start with just 2. (See also https://github.com/packit-service/deployment/issues/41)

Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+COPY ./files/ansible/install-deps.yaml /src/files/ansible/install-deps.yaml
+
+WORKDIR "/src"
+
+RUN ls -l && pwd && ansible-playbook -vv -c local -i localhost, files/ansible/install-deps.yaml \
+    && dnf clean all
+
+COPY . /src
+
+RUN ansible-playbook -vv -c local -i localhost, ./files/ansible/httpd.yaml
```
jpopelka: Don't you want to run `files/ansible/recipe.yaml` instead? Or where else is the `recipe.yaml` used?

Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+EXPOSE 8443
+
+CMD ["/src/files/scripts/run_httpd.sh"]
```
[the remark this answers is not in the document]
sakalosj: fixed

files/ansible/recipe.yaml
```diff
@@ -0,0 +1,9 @@
+---
+- name: This is a recipe for how to cook with packit-service
```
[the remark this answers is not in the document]
sakalosj: fixed

Dockerfile
sakalosj: fixed

files/cfg/httpd-packit.conf
sakalosj: fixed

Makefile
```diff
@@ -1,6 +1,17 @@
...
 install-dependencies:
 	sudo dnf -y install python3-flask npm
 	npm install
 
 run:
-	FLASK_ENV=development python3 main.py
+	FLASK_ENV=development FLASK_APP=packit_dashboard.app flask run
+
+run-docker:
+	docker run -p 443:8443 -v $(DIR)/secrets:/secrets -i usercont/packit-dashboard:stg
```
jpopelka: Can you define the image name as a variable at the top of the `Makefile`:
```
IMAGE ?= usercont/packit-dashboard:stg
```

and then

```suggestion
	docker run -p 443:8443 -v $(DIR)/secrets:/secrets -i $(IMAGE)
```

Makefile
```diff
@@ -1,6 +1,17 @@
...
+
 install-dependencies:
 	sudo dnf -y install python3-flask npm
 	npm install
 
 run:
-	FLASK_ENV=development python3 main.py
+	FLASK_ENV=development FLASK_APP=packit_dashboard.app flask run
+
+run-docker:
```
jpopelka:
```suggestion
run-docker: build-stg
```

Makefile
```diff
@@ -1,6 +1,17 @@
...
+	docker run -p 443:8443 -v $(DIR)/secrets:/secrets -i usercont/packit-dashboard:stg
+
+build-stg:
+	docker build -t usercont/packit-dashboard:stg -f Dockerfile . --rm
```
jpopelka:
```suggestion
	docker build --rm -t $(IMAGE) -f Dockerfile .
```

Makefile
```diff
@@ -1,6 +1,17 @@
...
+
+push:
```
jpopelka:
```suggestion
push: build-stg
```

Makefile
```diff
@@ -1,6 +1,17 @@
...
+	docker push usercont/packit-dashboard:stg
```
jpopelka:
```suggestion
	docker push $(IMAGE)
```

Makefile
sakalosj: fixed

Makefile
sakalosj: fixed

Makefile
sakalosj: fixed

Makefile
sakalosj: fixed

Makefile
sakalosj: fixed

sakalosj:
finaly :)
thanks for cooperation :)

Issue #5, opened 2020-01-27
requests.exceptions.ConnectionError


dhodovsk:
Dashboard can be moody - The requests fail with Connection error sometimes: 
```
127.0.0.1 - - [27/Jan/2020 16:39:33] "GET /projects/?__debugger__=yes&cmd=resource&f=debugger.js HTTP/1.1" 200 -
127.0.0.1 - - [27/Jan/2020 16:39:34] "GET /projects/?__debugger__=yes&cmd=resource&f=ubuntu.ttf HTTP/1.1" 200 -
127.0.0.1 - - [27/Jan/2020 16:39:34] "GET /projects/?__debugger__=yes&cmd=resource&f=console.png HTTP/1.1" 200 -
127.0.0.1 - - [27/Jan/2020 16:39:34] "GET /projects/?__debugger__=yes&cmd=resource&f=console.png HTTP/1.1" 200 -
127.0.0.1 - - [27/Jan/2020 16:39:56] "GET /projects/ HTTP/1.1" 500 -
Traceback (most recent call last):
  File "/usr/lib/python3.7/site-packages/flask/app.py", line 2309, in __call__
    return self.wsgi_app(environ, start_response)
  File "/usr/lib/python3.7/site-packages/flask/app.py", line 2295, in wsgi_app
    response = self.handle_exception(e)
  File "/usr/lib/python3.7/site-packages/flask/app.py", line 1741, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/lib/python3.7/site-packages/flask/_compat.py", line 35, in reraise
    raise value
  File "/usr/lib/python3.7/site-packages/flask/app.py", line 2292, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/lib/python3.7/site-packages/flask/app.py", line 1815, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/lib/python3.7/site-packages/flask/app.py", line 1718, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/lib/python3.7/site-packages/flask/_compat.py", line 35, in reraise
    raise value
  File "/usr/lib/python3.7/site-packages/flask/app.py", line 1813, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/lib/python3.7/site-packages/flask/app.py", line 1799, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
[...]

PR #9, opened 2020-02-23
Add Copr Builds to the dashboard.

IceWreck:
I added recent copr builds to the dashboard. Right now it only shows the recent ones because fetching all of them triggers issue #5 
I plan to add a button to load older builds as well so that it doesn't fetch too much data in one go.

The table can be sorted and filtered by project name, build status, etc thanks to datatables. (no additional dependencies, patternfly already has it as dependency)

Screenshot_2020-02-24 Screenshot(2)


sakalosj:
looks nice :+1:

templates/builds.html
```diff
@@ -0,0 +1,75 @@
...
+    {% for build in builds_list %}
+    <tr role="row">
+      <td role="cell" data-label="Project">{{ build["project"] }}</td>
+      <td role="cell" data-label="Owner">{{ build["owner"] }}</td>
+      <td role="cell" data-label="Status">{{ build["status"] }}</td>
+      <td role="cell" data-label="Chroots">
+        {% for os in build["chroots"] %}
+        <li>{{ os }}</li>
+        {% endfor %}
+      </td>
```
lachmanfrantisek: Sorting with the `target` does not make sense -- is it possible to turn it off or use some checkboxes? (We can improve that later.)

templates/builds.html
```diff
@@ -0,0 +1,75 @@
...
+        </button>
+      </th>
+
+    </tr>
+  </thead>
+  <tbody role="rowgroup">
+
+
+    {% for build in builds_list %}
+    <tr role="row">
```
lachmanfrantisek: Can we add some date+time to be able to get the original order easily? (And also to have that info?)

templates/builds.html
IceWreck: Yeah sure. I'll disable it for chroots

templates/builds.html
IceWreck: Hey, thanks for the review!

I may be wrong but I think as of now the packet-service api does not provide the build data/time so we should create a new issue in the [packet-service repo](https://github.com/packit-service/packit-service) for it.
Same goes for the link to the PR that triggers a new build.
I guess we can add a link to the repo of concerning project but even that has limited use considering that many builds do not provide a build id and without that we cannot make a request for the repo link.

IceWreck:
@lachmanfrantisek 
Like you suggested, I disabled the filter on the _build target OS_ column.

IceWreck:
@lachmanfrantisek I implemented the Pull Request link button which shows up only when the build id is provided. However, in some cases api/copr-builds/{id} returns nothing even when api/copr-builds/ provides the exact same id.

One solution would be to show the _View PR_ button on builds that do return an id and then in the rare case that the api returns nothing, just display an error message. But the proper way would be the fix the root cause of the issue, in the API code. 

So should I continue with the former solution or just leave it for now?

templates/builds.html
lachmanfrantisek: Please, create an issue for that.

We are currently switching to the "real" database, and we can have there what we need.
..;)

- [ ] Data needs to be saved in the database.
- [ ] The relevant info needs to be accessible via API.

IceWreck: @lachmanfrantisek Created packit-service/packit-service#425

lachmanfrantisek:
> One solution would be to show the _View PR_ button on builds that do return an id and then in the rare case that the api returns nothing, just display an error message. But the proper way would be the fix the root cause of the issue, in the API code.

As I wrote in the issue, these are probably the old data from early dates of the service.
- Should be completely fixed after we fully switch to PostgreSQL.
- Should not occur for new jobs.

> So should I continue with the former solution or just leave it for now?

Let's leave it as is and prepare the needed/wanted data in the API first.

IceWreck:
@lachmanfrantisek Rebased.

lachmanfrantisek:
> @lachmanfrantisek Rebased.

Thanks, merging.

PR #20, opened 2020-03-04
Display a list of projects instead of the raw python object

IceWreck:
- Replaced the raw object in the projects tab with a list. 
- I also ran black on the entire repo so some code was reformatted.
- Replaced the hardcoded word "information" (which was shown in every tab) with respective headers for each tab.

Screenshot_2020-03-04 Screenshot

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,14 +1,11 @@
 {
-  "name": "dashboard",
-  "version": "0.0.1",
-  "dependencies": {
-    "@patternfly/patternfly": "^2.46.1",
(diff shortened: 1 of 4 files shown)


jpopelka:
We don't have any contribution guide in this repo yet so I can't blame you, but this an example of where it would be nice to have the PR split in more commits.
At least have a separate commit for the black-reformatted changes.

I'd do `git reset HEAD^` and then `git add -p && git commit` to cherry-pick individual changes into separate commits. And finally `git push -f`.

Issue #21, opened 2020-03-05
Setup tests


lachmanfrantisek:
- [ ] Create some tests.
- [ ] Setup the directory structure as we have in other projects.
- [ ] Add Makefile target for that.
- [ ] Make the tests in a way we can easily run them in the CI.

lachmanfrantisek:
@IceWreck do you wantto work on it?

IceWreck:
Yes

lachmanfrantisek:
@IceWreck that was qucik..;)

IceWreck:
> @IceWreck that was quick..;)

Yeah I had my email open for homework and saw your message.

PR #24, opened 2020-03-06
Modularize project with flask blueprints

IceWreck:
So before I start working on #21 we need a proper project structure. We can't just keep adding code to one python file. So I used flask blueprints to split the app into blueprints (which are also subpackages).


IceWreck:
> nice! ready for merge?

@jpopelka Are you asking me? 
If so, yes. This is it for modularizing.

PR #26, opened 2020-03-10
modifications for pre-commit pass

rishavanand:
includes manual modifications by me and auto modifications by `pre-commit run --all-files`


rishavanand:
Kindly verify if `import os` is needed in `packit_dashboard/packit_dashboard.wsgi`. I had to remove it as the error thrown was `os imported but not used`.

packit_dashboard/packit_dashboard.wsgi
```diff
@@ -1,4 +1,4 @@
-import sys, os
+import sys
```
lachmanfrantisek: :+1: If it is not imported than it is not needed.

.pre-commit-config.yaml
```diff
@@ -27,13 +27,15 @@ repos:
       - id: jinjalint
         types: [file]
         files: \.(html)$
+        args:
+          - --parse-only
```
lachmanfrantisek: Without that, `jinjalint` will fix those files? If yes, I would remove it so we have the same approach as we have for black.

```sh
git commit
# oh, pre-commit fail because of some black problems
# black already fix those problems, we only need to add them:
git add
git commit
# no black passes
```

rishavanand: `jijalint` only checks for errors, it doesn't fix them. 
Yes, with indentation check disabled, it will still check for logical errors. 
The above case won't happen as fixing needs to be done manually.

csomh: > Yes, with indentation check disabled, it will still check for logical errors.

To me it seems that `--parse-only` as the name says, disables all checks, not only the indentation ones. I couldn't find any option to disable indentation only checks. I've tried this by invalidating one of the templates and trying to make a commit with this change. `jinjalint` happily reported `Passed` with this configuration.

rishavanand: I actually tried it already and it did throw the logical error. I'll recheck and report.

rishavanand: Here, try adding the following to one of the templates and then run `pre-commit`
```
<h2>
  test {% if false %}
</h2>
{% endif %}
```

Here is the error that was thrown on my side: 
```
templates/projects.html:49:24: Parse error: expected one of '[:a-z]', 'area', 'base', 'br', 'col', 'command', 'embed', 'hr', 'img', 'input', 'keygen', 'link', 'meta', 'param', 'script', 'source', 'style', 'track', 'wbr', '{#', '{%', '{{' at 48:24
```

csomh: :man_facepalming: 

Sorry @rishavanand, yesterday was a bad day for science for me.

`--parse-only` will indeed check syntax errors in jinja templating pieces, and my reasoning above does not make any sense :disappointed:

rishavanand: No worries. I keep having such days very often 😄

PR #40, opened 2020-06-08
Changed configuration to use wildcard certificates.

sakalosj:
Changed configuration to use wildcard certificates.


sakalosj:
> Please also update CONTRIBUTING.md, since the dashboard-* names are used in there.

fixed

PR #43, opened 2020-06-24
Jobs Page

IceWreck:
[Screenshare](https://cloud.abifog.com/f/dc4f8d950e52459f89ea/)

### DIFF
--- a/frontend/components/copr_builds_table.js
+++ b/frontend/components/copr_builds_table.js
@@ -55,11 +55,11 @@ const StatusLabel = (props) => {
 const CoprBuildsTable = () => {
     // Headings
     const column_list = [
-        { title: "Project", transforms: [sortable] },
+        "PR",
         "Chroots",
-        { title: "Build ID", transforms: [sortable] },
         { title: "Time Submitted", transforms: [sortable] },
-        "Ref",
+        { title: "Build ID", transforms: [sortable] },
+        // "Ref",
     ];
 
     // Local State
@@ -107,9 +107,9 @@ const CoprBuildsTable = () => {
                             <StatusLabel list={copr_builds.status_per_chroot} />
                         ),
                     },
-                    copr_builds.build_id,
                     copr_builds.build_submitted_time,
-                    copr_builds.ref,
+                    copr_builds.build_id,
+                    // copr_builds.ref.substring(0, 8),
                 ],
             };
             rowsList.push(singleRow);
--- a/frontend/components/copr_builds_table.js
+++ b/frontend/components/copr_builds_table.js
@@ -1,0 +1,187 @@
+import React, { useState, useEffect } from "react";
+
+import {
+    Table,
+    TableHeader,
+    TableBody,
+    TableVariant,
+    sortable,
+    SortByDirection,
+} from "@patternfly/react-table";
+
+import { Button, Label, Tooltip } from "@patternfly/react-core";
+import { ExternalLinkAltIcon } from "@patternfly/react-icons";
+
+import ConnectionError from "./error";
+import Preloader from "./preloader";
+
+// Add every target to the chroots column and color code according to status
+const StatusLabel = (props) => {
+    let statusLabelList = [];
+    let chroot;
+    for (chroot in props.list) {
+        switch (props.list[chroot]) {
+            case "success":
+                statusLabelList.push(
+                    <Tooltip content={props.list[chroot]}>
+                        <div style={{ padding: "2px" }}>
+                            <Label color="green">{chroot}</Label>
+                        </div>
+                    </Tooltip>
+                );
+                break;
+            case "failure":
+                statusLabelList.push(
+                    <Tooltip content={props.list[chroot]}>
+                        <div style={{ padding: "2px" }}>
+                            <Label color="red">{chroot}</Label>
+                        </div>
+                    </Tooltip>
+                );
+                break;
+            default:
+                statusLabelList.push(
+                    <Tooltip content={props.list[chroot]}>
+                        <div style={{ padding: "2px" }}>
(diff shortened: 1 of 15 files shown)


lachmanfrantisek:
> Screenshare

Thanks for making that! It's looking very neat!

TomasTomecek:
are you kidding me?! this looks so awesome!

frontend/components/copr_builds_table.js
```diff
@@ -0,0 +1,215 @@
...
+// Add every target to the chroots column and color code according to status
+const StatusLabel = (props) => {
+    let statusLabelList = [];
+    let chroot;
+    for (chroot in props.list) {
+        switch (props.list[chroot]) {
+            case "success":
+                statusLabelList.push(
+                    <div style={{ padding: "2px" }}>
+                        <Label color="green">{chroot} - Success</Label>
```
csomh, line 42: Would it be possible to remove "Success" from here, either by moving it to some HTML accessibility attribute, or just making it hidden?

frontend/components/copr_builds_table.js
```diff
@@ -0,0 +1,215 @@
...
+    // Headings
+    const column_list = [
+        "Namespace",
+        { title: "Repository", transforms: [sortable] },
+        { title: "Build ID", transforms: [sortable] },
+        { title: "Time Submitted", transforms: [sortable] },
+        "Chroots",
+        // "Ref",
+        "Web URL",
+        { title: "PR ID", transforms: [sortable] },
```
csomh, line 76: As discussed, would be cool if this could be reworked as follows:

PR | Chroots | Time Submitted | Build ID
------|----|---|----
namespace/repository#prid (and linking to weburl) | chroots | time | build id

frontend/components/copr_builds_table.js
```diff
@@ -0,0 +1,215 @@
...
+        return (
+            <div>
+                <center>
+                    <br />
+                    <Spinner size="xl" />
+                    <br />
+                </center>
+            </div>
+        );
+    }
```
csomh, line 190: If possible factor out these two blocks as they show up in the Testing Farm result page too.

packit_dashboard/api/routes.py
```diff
@@ -0,0 +1,26 @@
+from flask import Blueprint, jsonify, request
+from packit_dashboard.utils import return_json
+from packit_dashboard.config import API_URL
+
+api = Blueprint("api", __name__)
+
+# The react frontend will request information here instead of fetching directly
+# from the main API.
+# This is because it will be easier to implement caching API requests here.
+# (Flask-Caching etc)
```
csomh, line 10: I feel that doing caching in the API would be a more robust solution on the long run. For now, let's keep it this way, though.

IceWreck: Yeah. However, the downside of caching the API in packit-service is that we won't be able to get new uncached info even if we want to

frontend/components/copr_builds_table.js
```diff
@@ -0,0 +1,187 @@
...
+
+    // Convert fetched json into row format that the table can read
+    function jsonToRow(res) {
+        let rowsList = [];
+        res.map((copr_builds) => {
+            let singleRow = {
+                cells: [
+                    {
+                        title: (
```
csomh, line 96: Why is using `title` here and in the next field required?

frontend/components/copr_builds_table.js
```diff
@@ -0,0 +1,187 @@
...
+                );
+        }
+    }
+    return <div>{statusLabelList}</div>;
+};
+
+const CoprBuildsTable = () => {
+    // Headings
+    const column_list = [
+        { title: "Project", transforms: [sortable] },
```
csomh, line 58:
```suggestion
        { title: "PR", transforms: [sortable] },
```

frontend/components/copr_builds_table.js
```diff
@@ -0,0 +1,187 @@
...
+        "Chroots",
+        { title: "Build ID", transforms: [sortable] },
+        { title: "Time Submitted", transforms: [sortable] },
```
csomh, line 61: Reverse:

```suggestion
        { title: "Time Submitted", transforms: [sortable] },
        { title: "Build ID", transforms: [sortable] },
```

frontend/components/copr_builds_table.js
```diff
@@ -0,0 +1,187 @@
...
+        "Ref",
```
csomh, line 62: Isn't this going to take up too much space? I would suggest to remove it for now in order to keep the table simple and clean.

IceWreck: It takes around a quarter of the total space but since we merged Namespace/Repo#PR and removed Success/Failure in favour of color coding + tooltips we had a lot of empty space left so I added ref.
Screenshot_2020-06-26 Packit Service

frontend/components/copr_builds_table.js
IceWreck: Let me check...

frontend/components/copr_builds_table.js
lachmanfrantisek: What about having only the first `n` characters? (E.g. with `git` commits, it's common to use only first `8` characters.)

But we can have also long branch names here, don't we? So use some prefix with elypsis for long refs?

csomh: Empty space is fine. Let's remove "Ref" for now and we can improve this screen in future iterations if we want.

In general the approach we discussed was to keep this list minimal and maybe have the rows expand on click in the future, displaying more information about the individual jobs.

frontend/components/copr_builds_table.js
csomh: Okay, thanks :+1:

PR #44, opened 2020-06-26
Cache API requests and wait before retrying

IceWreck:
Packit Service API fails to respond when faced with concurrent requests. We should fix this issue there too but we don't know the exact cause

In the meantime, caching and retrying here helps.

### DIFF
--- a/packit_dashboard/cache.py
+++ b/packit_dashboard/cache.py
@@ -1,0 +1,9 @@
+from flask_caching import Cache
+
+cache = Cache(
+    config={
+        "CACHE_TYPE": "filesystem",
+        "CACHE_DEFAULT_TIMEOUT": 120,
+        "CACHE_DIR": "/tmp",
+    }
+)
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ TEST_TARGET ?= ./tests/
 CONTAINER_ENGINE ?= $(shell command -v podman 2> /dev/null || echo docker)
 
 install-dependencies:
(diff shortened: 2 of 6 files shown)


IceWreck:
Actually I have another idea that can work alongside this.

Right now both API requests are made simultaneously when the user loads the dashboard which can occasionally cause PS to crash. We can simply load the other tab (testing farm results) when we click on it the first time instead of loading it in the background.

packit_dashboard/cache.py
```diff
@@ -0,0 +1,9 @@
+from flask_caching import Cache
+
+cache = Cache(
+    config={
+        "CACHE_TYPE": "filesystem",
```
TomasTomecek, line 5: could we cache in redis/psql? that should be faster

IceWreck:
> Is there any reason we couldn't use the same mechanism on Packit Service side? thinking I would say even a 5 seconds cache on these expensive API endpoints would help stabilising things.

Sure, this should work there as well. 
Its just that caching it here would mean we can cache for longer durations if we want but doing that there would mean we cant get fresh data even if we want to

packit_dashboard/cache.py
IceWreck: Yes, but isnt that too overcomplicated for four or five api requests ?

csomh:
I think solving packit-service/packit-service#706 could make caching unnecessary. At least for now :smiley:

IceWreck:
Closing this for now.

PR #46, opened 2020-07-14
Projects

IceWreck:
[Screenshare](https://cloud.abifog.com/f/fc6f307f515142ee81d6/)

### DIFF
--- a/frontend/components/pr.js
+++ b/frontend/components/pr.js
@@ -1,0 +1,277 @@
+import React, { useState, useEffect } from "react";
+
+import ConnectionError from "./error";
+import Preloader from "./preloader";
+
+import {
+    Button,
+    DataList,
+    DataListToggle,
+    DataListCell,
+    DataListItem,
+    DataListContent,
+    DataListItemCells,
+    DataListItemRow,
+    Dropdown,
+    DropdownToggle,
+    DropdownItem,
+    Label,
+} from "@patternfly/react-core";
+
+import { CaretDownIcon, ExternalLinkAltIcon } from "@patternfly/react-icons";
+
+const PullRequestList = (props) => {
+    // Local State
+    const [hasError, setErrors] = useState(false);
+    const [loaded, setLoaded] = useState(false);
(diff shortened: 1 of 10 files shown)


frontend/components/pr.js
```diff
@@ -0,0 +1,277 @@
...
+    // Open/close the dropdown
+    // This is called in two cases
+    // a) when the toggle button is pressed
+    // b) when someone chooses an entry
+    function onToggle() {
+        setOpen((isOpen) => !isOpen);
+    }
+
+    // NOTE: Since we did not need any advanced table features,
+    // I used a regular html table instead of Patternfly React's table component
```
csomh, line 141: :+1:

frontend/components/releases.js
```diff
@@ -0,0 +1,83 @@
...
+            .then((response) => response.json())
+            .then((data) => {
+                setReleaseList(data);
+                setLoaded(true);
+            })
+            .catch((err) => {
+                console.log(err);
+                setErrors(err);
+            });
+    }
```
csomh, line 31: This seems to be a little repetitive :thinking:

packit_dashboard/utils.py
```diff
@@ -13,7 +13,7 @@ def return_json(url, method="GET", **kwargs):
         try:
             response = requests.request(method=method, url=url, **kwargs)
             output = json.loads(response.content)
-            print(f"Try-{i}")
+            # print(f"Try-{i}")
```
csomh, line 16: Is this line needed?

frontend/components/releases.js
IceWreck: Yes this comes up in every file but we cant abstract this further because the url scheme, and the then {} part is different is at best we will save two or three lines of code.

packit_dashboard/utils.py
IceWreck: Not really, I should remove this

IceWreck:
Screenies of the searchbar
image

and 
image

PR #49, opened 2020-07-26
Unrelated Misc Changes

IceWreck:
- [x] Interchange git url and project page links
- [x] end of data (after clicking load more) should not result it error - (i think this is fixed in packit-service/packit-service/pull/754 and no change needs to be done here, but my current deployment of ps mirrors an old prod dump so i dont have enough patience to click load more 100 times to verify )
- [x] wildcard search for projects via forge + namespace
- [x] branches handled view
- [x] fix the name scheme in jobs/copr-builds in cases where its a branch push instaed of pull request

Note: To be merged only after https://github.com/packit-service/packit-service/pull/754


frontend/components/copr_builds_table.js
```diff
@@ -90,6 +90,18 @@ const CoprBuildsTable = () => {
...
     function jsonToRow(res) {
         let rowsList = [];
+
+        // set suffix to be either PR ID or Branch Name depending on trigger
+        const FindSuffix = (props) => {
+            let jobSuffix = "";
+            if (props.builds.pr_id) {
+                jobSuffix = `#${props.builds.pr_id}`;
+            } else if (props.builds.branch_name) {
+                jobSuffix = `-${props.builds.branch_name}`;
```
csomh, line 100: GitHub (and maybe also other places) use `:` to separate the branch name from the project name. Could we be consistent with that usage? (furthermore `-` could be misleading when the project name contains `-`.)

frontend/components/copr_builds_table.js
```diff
@@ -90,6 +90,18 @@ const CoprBuildsTable = () => {
...
+            }
+            return <>{jobSuffix}</>;
```
csomh, line 102: Hm, why is "<> ... </>" needed here? It looks very strange :eyes:

frontend/components/copr_builds_table.js
IceWreck: 👍

templates/index.html
```diff
@@ -24,6 +24,9 @@
     <body class="pf-m-redhat-font">
         <noscript>Enabling JavaScript is required to run this app.</noscript>
         <div id="root"></div>
+        <script>
+            window.apiURL = "{{api_url}}";
```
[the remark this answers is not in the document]
IceWreck: window.variableName is a bad practice for global variables in JS. And you're right, const apiURl = ... is working here so why use this. Will change this.

frontend/components/copr_builds_table.js
IceWreck: React needs something to wrap elements at the top of the tree. I generally use <div> or <span> but if there is no need for those we can use <>..</>

templates/index.html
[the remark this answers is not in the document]
csomh: I see you updated this, but did not use `const` as you suggested above. Was there any reason for it?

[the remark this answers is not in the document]
IceWreck: I remember adding it, dunno how it disappeared. Seriously! 
And JavaScript being the language that it is, 😠😠😠 did not warn that `const/let/va`r was missing and continued to work.

PR #60, opened 2020-10-29
Rectify koji builds status tooltip text

IceWreck:
Earlier, the tooltip displayed the chroot which was already mentioned in the label. Meanwhile the status was only reflected by the color.
This is how it looked:
image

Now the tooltip shows the status.


jpopelka (approved):
:+1:

Issue #63, opened 2020-11-04
Ignore projects without any handled item


lachmanfrantisek:
Most of the project on the projects page is irrelevant because they don't handle any pr/issue/branch/release.

What about skipping those projects (they don't bring any value)? Or adding some option to opt-in/opt-out from listing them.

IceWreck:
I see two options here,
a) handling this in the query in service
b) hiding the projects which have 0 PRs && 0 Branches handled

the second one would be easier and would allow us to add a toggle to opt-in/opt-out from viewing them

lachmanfrantisek:
I am for the second one because of the reason you described.

Just a detail, I would count also issues and releases (so all four numbers we show have to be zero).

off-topic: It might sound picky from me to create multiple issues here, but I would like to say it's really cool that we have the dashboard! (I needed to contact our users and was glad that I can use dashboard instead of the database..;)

lachmanfrantisek:
Thanks @IceWreck !

lachmanfrantisek:
This is solved, isn't it?

Only the last point (`Add toggle to view ignored projects (later)`) is missing and will be done later.

PR #64, opened 2020-11-09
Ignore projects without any handled item

IceWreck:
Solves #63 

TODO:

- [x] Ignore projects without any handled item 
- [x] fetch more projects if useful projects <= 5 (out of 50) 
- [ ] Add toggle to view ignored projects (later)

### DIFF
--- a/frontend/components/projects_list.js
+++ b/frontend/components/projects_list.js
@@ -29,6 +29,8 @@ const ProjectsList = (props) => {
     const [loaded, setLoaded] = useState(false);
     const [page, setPage] = useState(1);
     const [projects, setProjects] = useState([]);
+    // by default, ignore projects without any handled item
+    const [showUseful, setShowUseful] = useState(true);
 
     // If a namespace and forge are provided, then load those
     // otherwise load all projects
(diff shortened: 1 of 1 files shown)


lachmanfrantisek:
Looks good, thanks!

Maybe an odd question but will this work with the lazy loading of the new projects?

I am just thinking about the case when we load a new group of projects where all the projects will be skipped. It will do nothing in that case, won't it? (There are many of those projects so it is possible.)

IceWreck:
> I am just thinking about the case when we load a new group of projects where all the projects will be skipped. It will do nothing in that case, won't it? (There are many of those projects so it is possible.)


If we click load more and it fetches 50 new projects and lets say 40 of those are ignored, then it will only display 10. On loading more, it will fetch the next 50 projects and apply the same logic to them.

In case all 50 are ignored, then yes it will display nothing and you'll have to click load more again to view the next batch.

IceWreck:
> In case all 50 are ignored, then yes it will display nothing new and you'll have to click load more again to view the next batch.

This could be fixed by giving some kind of indication about the current page at the bottom. Like page numbers or something. So that will give an indication that the page has changed even if there are no new visible cards. And once the API starts returning an empty list, we can tell the user that theyve reached the last page

lachmanfrantisek:
It can easily get to the situation that you need to click multiple times without any response to get new values. (There is a user in the middle with a really big amount of ignored packages.)

lachmanfrantisek:
Can we load another group when we ignore all 50 packages?

IceWreck:
> Can we load another group when we ignore all 50 packages?

I spent multiple hours trying to make it work but there is no way to reliably predict how many times the react component is rerendered and when it has finished rendering. That and fighting async/await made me give up for today.

Basically I was deciding what to ignore **just before render.** This works if I'm just ignoring/allowing them but if I want to keep track of how many total projects ive ignored I'll have to maintain a list in the state. 

This should be a trivial one liner change but isnt cause there is no such thing as 'append to DOM' in react, the entire component is re-rendered whenever the state changes (ie whenever the projects list is modified)

Its also possible that I'm making a very very stupid mistake, but I think the proper way to do it is to just decide what to ignore after fetching it, not deciding just before render

I'll do it properly tomorrow or someday this week.

lachmanfrantisek:
@IceWreck thanks for spending so much time on it!

IceWreck:
Uggghhh Done !!!!!!!

I spent so much time overcomplicating things including:

- adding babel plugins so i could use JavaScript ES7 features like async/await
- asynchronous recursion in js which didnt sit well react useState
- trying to keep track of react rerenders so id know when to fetch more  
- unsuccessful time based hacks (just to try them out, i know stuff like this will never be used in production)

and finally I re-read useEffect is the React docs and the solution was so damn simple.

IceWreck:
BTW you should consider removing the github user `m4ta1l` from the prod db. They deleted or renamed their GitHub account and have enabled packit service in 2744 repos out of which only two are actually using packit.

While after this PR the dashboard will ignore their unused repos, it still takes 50-70 seconds to automatically load the next useful repository. So we have to add some kind of indication (other than console logs) that its loading more repos (in a separate PR, not this one please)

lachmanfrantisek:
> BTW you should consider removing the github user `m4ta1l` from the prod db. They deleted or renamed their GitHub account and have enabled packit service in 2744 repos out of which only two are actually using packit (ogr fork and one other).

Yes, this user was the reason I've created that issue..;)

lachmanfrantisek:
> Uggghhh Done !!!!!!!

Nice that you've figured that out!

lachmanfrantisek:
@IceWreck Are we missing something? Do you consider this still a WIP?

I am really looking forward to seeing this merged..;)

IceWreck:
> @IceWreck Are we missing something? Do you consider this still a WIP?
> 
> I am really looking forward to seeing this merged..;)

Need to add a button to view ignored ones as well. I have daily quizzes till 15th Dec instead of exams so Ill do it after that. 
Other than the button its done. If you want I'll rebase this now so it can be merged and then add the button (and some indication about the current page) in a separate PR.

lachmanfrantisek:
> If you want I'll rebase this now so it can be merged and then add the button (and some indication about the current page) in a separate PR.

Yes, that would be nice!

Good luck with the quizzes!

lachmanfrantisek (approved):

lachmanfrantisek:
Thank you!

csomh (approved):
:ice_cube: :rocket:

PR #84, opened 2021-03-26
Show SRPM link only if present

mfocko:
Signed-off-by: Matej Focko <mfocko@redhat.com>

### DIFF
--- a/frontend/components/results_srpm.js
+++ b/frontend/components/results_srpm.js
@@ -61,7 +61,7 @@ const ResultsPageSRPM = (props) => {
     }
 
     const srpmURL = data.url ? (
-        <a href={data.url}>Download</a>
+        <a href={data.url}>Link to download</a>
     ) : (
         "Not available to download"
     );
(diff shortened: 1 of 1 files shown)


frontend/components/results_srpm.js
```diff
@@ -60,6 +60,14 @@ const ResultsPageSRPM = (props) => {
         );
     }
 
+    const srpmURL = data.url ? (
+        <strong>
+            <a href={data.url}>Downloadable SRPM</a>
+        </strong>
+    ) : (
+        <a></a>
```
TomasTomecek, line 68: I'm wondering if we could make the else branch better, maybe "SRPM not available yet"?

mfocko: I'm actually thinking about moving it up (above the logs)

mfocko:
Looks something like this at the latest commit:
image

TomasTomecek (approved):
a neat improvement!

PR #89, opened 2021-04-01
COPR builds improvements

mfocko:
**Do not merge before packit/packit-service#1040**

### DIFF
--- a/frontend/components/copr_builds_table.js
+++ b/frontend/components/copr_builds_table.js
@@ -17,40 +17,36 @@ import Preloader from "./preloader";
 import ForgeIcon from "./forge_icon";
 
 // Add every target to the chroots column and color code according to status
-const StatusLabel = (props) => {
-    let statusLabelList = [];
-    let chroot;
-    for (chroot in props.list) {
-        switch (props.list[chroot]) {
+const ChrootStatuses = (props) => {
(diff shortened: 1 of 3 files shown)


frontend/components/copr_builds_table.js
```diff
@@ -109,7 +104,10 @@ const CoprBuildsTable = () => {
                     },
                     {
                         title: (
-                            <StatusLabel list={copr_builds.status_per_chroot} />
+                            <ChrootStatuses
+                                statuses={copr_builds.status_per_chroot}
+                                ids={copr_builds.packit_id_per_chroot}
```
mfocko, line 109: I guess it would be nice to return:
```json
"chroots": {
    "fedora-33-x86_64": {
        "id": 1234,
        "status": "success"
    }
}
```

TomasTomecek: yes, it would: but since we are the only users of the API, I wouldn't worry too much about it :)

TomasTomecek (approved):

frontend/components/copr_builds_table.js
mfocko: I might look into it, cause we're starting to grow indexing in dashboard because of this O.o

PR #122, opened 2021-07-19
Allow filtering projects of a given forge

lbarcziova:


frontend/src/components/search_project.js
```diff
@@ -33,7 +33,13 @@ const SearchProject = () => {
                 history.push(`/projects/${forge}/${namespace}/${repoName}`);
             }
         } else if (forge && namespace) {
-            history.push(`/projects/${forge}/${namespace}`);
+            if (namespace === "*") {
+                history.push(`/projects/${forge}`);
```
lachmanfrantisek, line 37: Can we use empty value as all?

Thinking about the `*`, wouldn't it lead to an illusion that we support regexes?

packit_dashboard/api/routes.py
```diff
@@ -35,6 +35,14 @@ def projects():
     return jsonify(return_json(url))
 
 
+@api.route("/api/projects/<forge>/")
+def projects_forge(forge):
```
lachmanfrantisek, line 39: :+1:

lachmanfrantisek (commented):
Thanks a lot!

frontend/src/components/search_project.js
lbarcziova: Currently, it works both for empty value and `*`, following the behaviour for getting projects of a namespace. Should we then also remove the possibility for getting projects of a namespace by `*` in the repository field?
Since in the fields, there are the hints `the-repo-name-or-*` and `the-namespace-or-*`, I think it should be clear that we don't support regexes, on the other hand, it should be enough to leave a blank field for getting everything. But then we should probably somehow mark that the namespace and repo fields are optional, WDYT?

IceWreck:
Hey @lbarcziova 

I just wanted to point out that the changes in `packit_dashboard/api/routes.py` are unnecessary. We were using flask to make the actual API queries and then send them over to the frontend when 
* API requests were failing and we thought it required caching.
* we needed a temporary way to fix CORS issues till we added the appropriate headers everywhere in Packit Service.

Since those have been fixed, this is no longer required or used.

image

Your code here is directly calling packit service API, not `dashboardurl/api/projectswhatever`

frontend/src/components/search_project.js
lachmanfrantisek: If we don't want to support regexes, I would try to get rid of `*` and mark the fields as optional if it is possible.

Thanks!

lachmanfrantisek (approved):
I haven't tested it myself, but looks good to me.

lbarcziova:
> I haven't tested it myself, but looks good to me.

Thanks. I just changed the warning since I realised while testing that it doesn't make sense to output `Forge field should not be empty` when e.g.only forge and repo name are filled.

TomasTomecek:
Is this available on stg already?

lbarcziova:
@TomasTomecek yes: https://dashboard.stg.packit.dev/projects/github.com

lachmanfrantisek:
Sadly, we have the wrong icon for `gitlab.com`:
https://dashboard.stg.packit.dev/projects/gitlab.com

Screenshot from 2021-07-21 13-25-50

mfocko:
@lachmanfrantisek  Yeah, I noticed, will fix it soon; also looking into the stage icon ;)

lbarcziova:
We don't differentiate the icons for forges and use everywhere the github one currently, https://dashboard.packit.dev/projects/gitlab.com/lachmanfrantisek, https://dashboard.packit.dev/projects/gitlab.com/packit-service/hello-world, so it should be then changed there as well.

PR #128, opened 2021-07-22
Status labels rework + pipelines view

mfocko:
TODO:
- [x] revert the links on status labels in jobs tables
  - [x] decide between adding column or keeping direct link to external service in the result view
- [x] pipelines view
  - [x] refactor
  - [x] document
  - [x] _pushee pushee_
- [x] create issues for the potential bugs in service
  - [x] Koji builds have separate SRPM
    **intended**
  - [x] TF retrigger might be run from wrong COPR build
    packit/packit-service#1174

---

*release notes will go here*


mfocko:
I would like some opinions on:

1. Jobs tables
   1. SRPM - points to result page and logs on dashboard
   1. COPR - points to result page on dashboard
   1. Koji - points to Koji URL
   1. TF   - points to testing farm URL
1. Result pages
   1. SRPM - not interactive
   1. COPR - points to COPR URL
   1. Koji - points to Koji URL
   1. TF   - points to testing farm URL

It still seems like a mess to me, cause it's inconsistent, but for COPR we benefit from it since we are able to link all chroots to our own result page

lachmanfrantisek:
I would try to use the dashboard links where possible so users are able to go through the dashboard easily. (So, provide external links only in the result views.)

Regarding the mess, I agree with being consistent.

And from the user perspective, the merging of the Copr/Koji chroots in the job table looks good.

mfocko:
> I would try to use the dashboard links where possible so users are able to go through the dashboard easily. (So, provide external links only in the result views.)

I will try to think about that.

> Regarding the mess, I agree with being consistent.
> 
> And from the user perspective, the merging of the Copr/Koji chroots in the job table looks good.

Yeah, that's kind of a bummer. We cannot do it **correctly** apart from COPR, where chroots have same build ID, but different IDs on our side. Neither Koji or TF can be decided that easily... Even if we go by runs/pipelines, we still can't tell apart re-triggered runs, which should be currently possible, since we allow only re-triggering tests  or whole pipelines, but if we support rerun from GitHub Action, it could become a problem. Maybe it is time to implement view for pipelines that would simplify it :man_shrugging:

TomasTomecek:
What I really love about the copr page is that there **2** links: one to webui with results and one **directly** to logs.

Would be awesome to have the same for others but I understand it's close to impossible.

Agree with Franta on having a dashboard page for all jobs.

mfocko:
umm, _I've been cooking a bit_ on Friday and changes are dependent on those, so I will blow this PR up :)

Sneak peek: 

I reproduced the issue I mentioned, right in the first row by:

```
/packit build
/packit test
```

Apart from those I have noticed few issues:
1. `/packit test` uses latest COPR build in case none is given (e.g. failed COPR build), but that can result in using build from totally different PR/branch/release
2. Koji is built from separately created SRPM? Is that intentional?

lachmanfrantisek:
>     2\. Koji is built from separately created SRPM? Is that intentional?

We don't store the SRPM anywhere and the SRPM build is inside of the koji/copr build job.

mfocko:
> > ```
> > 2\. Koji is built from separately created SRPM? Is that intentional?
> > ```
> 
> We don't store the SRPM anywhere and the SRPM build is inside of the koji/copr build job.

OK, so if COPR+Koji cannot be in the same pipeline, maybe we could do something like this?

mfocko:
tl;dr After discussion with @lachmanfrantisek I have ended up with multiple designs. Please choose the one you like the most.

Context: Pipelines are merged by SRPM build, SRPM build is unique for COPR and Koji even if they are both defined to be run. Therefore Koji and COPR form separate pipelines/runs even if triggered by the same event (comment, push, release), which creates dead space in the first design and label groups seem a bit non-intuitive to Franta. :)

Option :apple:


Option :cherries: 


Option :watermelon:

csomh:
I like both :rocket: and :eyes:. How about moving 'COPR'/'Koji' in a separate "Built with" column? I think it would make the view more structured.

mfocko:
@csomh You mean separate columns for COPR and Koji?

That's done in the first option, the problem is that runs are always like:

    SRPM → ‹multiple COPR builds› → ‹multiple TF runs›
    SRPM → ‹multiple Koji builds›

And since majority of builds are COPR, it means that we end up with column of dead space for Koji :/

I was actually considering merging all columns into one and do rows of SRPM, COPR/Koji, Testing Farm, but I think it might be worse…

mfocko:
@csomh or you meant something like this?

csomh:
> @csomh or you meant something like this?

Exactly :rocket: :slightly_smiling_face:

mfocko:
Current state:

mfocko:
Done from my side for now.

We have separate pipelines view. I will create follow-up issues for result page for single pipeline and breadcrumbs navigation through the builds/tests on pipeline from result pages, e.g. SRPM logs or Copr build on our side. I expect those to be finished for deployment on next Friday.

frontend/src/routes.js
```diff
@@ -29,6 +30,13 @@ const routes = [
         path: "/jobs",
         title: "Jobs | Packit Service",
     },
+    {
+        component: Pipelines,
+        exact: true,
+        label: "Pipelines (beta)",
+        path: "/pipelines",
+        title: "Pipelines (beta) | Packit Service",
```
lachmanfrantisek, line 38: Do we need to call it `beta`? But it might look new and users will click on it just because of it..;)

lachmanfrantisek (approved):
I can't review the js code, but we've discussed the result so LGTM from my side.

Thanks a lot for this huge improvement!

frontend/src/components/tf_results_table.js
```diff
@@ -16,37 +16,15 @@ import ConnectionError from "./error";
...
+import { TFStatusLabel } from "./status_labels";
 
 const TestingFarmResultsTable = () => {
     const column_list = [
         { title: "", transforms: [cellWidth(5)] }, // space for forge icon
         { title: "Trigger", transforms: [cellWidth(35)] },
         { title: "Target", transforms: [sortable, cellWidth(20)] },
         { title: "Time Submitted", transforms: [cellWidth(20)] },
-        { title: "Results", transforms: [cellWidth(20)] },
+        { title: "TF Results", transforms: [cellWidth(20)] },
```
csomh, line 27: Not sure about this, but wouldn't it be more clear to call this "Test Results"? I have the feeling that not everyone would now what "TF" is.

csomh (approved):

frontend/src/routes.js
mfocko: Up to you :) I would keep it as _beta_ at least till we decide to deprecate jobs tables and hide them from the dashboard.

PR #137, opened 2021-08-23
Log viewer

### LINKED ISSUE #120: Build logs should use use browser wrapping
At this moment the build logs are not wrapped to browser window size and that makes it quite hard to read essential information from lines that are longer than the window.

https://dashboard.packit.dev/results/srpm-builds/28981

This [css hack](https://stackoverflow.com/questions/248011/how-do-i-wrap-text-in-a-pre-tag) seems to achieve the desired wrapping:
```
code {
    white-space: pre-wrap;
}
```

This behavior would also match what a normal user terminal would see, as terminals do not usually hide the long lines and do wrap them instead.

mfocko:
Fixes #120

---

Dashboard now uses log viewer component that provides wrapping of lines, line numbering and also built-in search.
screenshot
screenshot with search results


frontend/package.json
```diff
@@ -15,6 +16,7 @@
     "react-dom": "^17.0.2",
     "react-router-dom": "^5.2.0",
     "react-scripts": "4.0.3",
+    "resize-observer-polyfill": "^1.5.1",
```
mfocko, line 19: That should be removed, once it gets included in the upstream of `@patternfly/react-log-viewer`

PR: patternfly/patternfly-react#6216

lachmanfrantisek (approved):
:tada:

PR #142, opened 2021-09-23
Datetime improvements

mfocko:
Merge with packit/packit-service#1223

---

Dashboard now shows more readable format of time, e.g. _just now_, _a minute ago_, etc. If you wish to see exact date-time of the run, you can either hover over the time and tooltip with details appear. In case of result pages more readable format is present in the tooltip rather than by default. Also times are now shown in your local time zone.

For tables: 
For results:


jpopelka (approved):

TomasTomecek (commented):
:heart:ing this!

PR #144, opened 2021-10-01
End pipelines' beta phase

mfocko:
---

We consider [Pipelines view](https://dashboard.packit.dev/pipelines) on dashboard to be ready for production use. We're planning to improve the user experience when using pipelines with easier navigation and filtering options.


jpopelka (approved):

TomasTomecek (approved):
:zap:

PR #163, opened 2022-01-25
Update views to reflect SRPM changes related to building in Copr

lbarcziova:
SRPMBuildModel in packit-service was updated, it has now status attribute instead
of success and additional Copr related attributes, these changes are reflected into
packit-service API and the info we process in dashboard, therefore the views
need to be updated to show correct status and new info.

Related to packit/packit-service#1324

For SRPMs built by us we will show also start and finished time:
Snímka obrazovky z 2022-01-25 12-49-50

For SRPMs built in Copr there are additional links:
Snímka obrazovky z 2022-01-25 12-47-53

---

### DIFF
--- a/frontend/src/components/results/srpm.js
+++ b/frontend/src/components/results/srpm.js
@@ -105,7 +105,6 @@ const ResultsPageSRPM = (props) => {
         ""
     );
 
-
     const logs = data.copr_build_id ? (
         ""
     ) : (
--- a/frontend/src/components/results/srpm.js
+++ b/frontend/src/components/results/srpm.js
@@ -71,47 +71,88 @@ const ResultsPageSRPM = (props) => {
         "Not available to download"
     );
 
+    const submittedAt = data.build_submitted_time ? (
+        <Timestamp stamp={data.build_submitted_time} verbose={true} />
+    ) : (
+        "not available"
+    );
+
+    const startedAt = data.build_start_time ? (
+        <Timestamp stamp={data.build_start_time} verbose={true} />
+    ) : (
+        "not available"
+    );
+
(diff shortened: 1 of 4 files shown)


frontend/src/components/results/srpm.js
```diff
@@ -71,47 +71,87 @@ const ResultsPageSRPM = (props) => {
         "Not available to download"
     );
 
+    const submittedAt = data.build_submitted_time ? (
+        <Timestamp stamp={data.build_submitted_time} verbose={true} />
+    ) : (
+        "not available"
+    );
```
lachmanfrantisek, line 78: What about skipping that instead of `not available`? What do you think? (I am not sure what is better..;)

mfocko: I would probably refactor it into separate components as Friday work :shrug: since that would the most clean solution imo

lbarcziova: I implemented it in a way that `not available` will be shown if the value will be later added (since all the new SRPM builds will have also start and finish time stored and we do this also for the link for downloading the SRPM) and then info like Copr web URL and Copr logs will be shown only for the builds built by Copr and text logs only for builds built by us. But I do not have a strong preference about this, so if you think it shouldn't be shown until the value is not present, I can reimplement it.

jpopelka (approved):

frontend/src/components/results/srpm.js
lachmanfrantisek: @lbarcziova your reasoning makes sense. Thanks.

lachmanfrantisek (approved):

PR #172, opened 2022-03-04
Update the test image to Fedora Linux 35

csomh:
From 31 o_O!!!!

Signed-off-by: Hunor Csomortáni <csomh@redhat.com>


jpopelka (approved):
> From 31 o_O!!!!

that's because we haven't run the tests since 45ece391  :-O

PR #178, opened 2022-05-06
Propose downstream pipelines

mfocko:
Preview:

---

RELEASE NOTES BEGIN
Dashboard now shows also `propose-downstream` jobs in pipelines view. We have also merged all jobs to one column to minimize dead space on the page.
RELEASE NOTES END


TomasTomecek (approved):
whaaaaat, love this!

PR #181, opened 2022-06-17
Use mod_wsgi-express to serve the dashboard

csomh:
Because "provides a much better default configuration if running
mod_wsgi inside of a container".

Also: use 'yarn' to install and build the frontend. This was already so in the Makefile, but the recipe still used npm.


csomh:
Note, that it's still not possible to upgrade the base image to Fedora Linux 36, as openssl 3.0 doesn't play well with NodeJS 16, and breaks building the frontend. Wasn't able to find a solution for this just yet, but I'll keep looking.

files/scripts/run_httpd.sh
```diff
@@ -1,12 +1,14 @@
...
 
-set -x
+set -eux
 
-# Generate passwd file based on current uid
-grep -v ^packit /etc/passwd > ${HOME}/passwd
-printf "packit:x:$(id -u):0:Packit Service:/home/packit_dashboard:/bin/bash\n" >> ${HOME}/passwd
-export LD_PRELOAD=libnss_wrapper.so
-export NSS_WRAPPER_PASSWD=${HOME}/passwd
-export NSS_WRAPPER_GROUP=/etc/group
```
jpopelka, line 10: Do we still need to [install nss_wrapper](https://github.com/packit/dashboard/blob/33e4e89f3ffaf542d2cafb6dcd234b99b3517401/files/ansible/install-deps.yaml#L14) then?

csomh: No, we [don't](https://github.com/packit/dashboard/pull/181/files#diff-c4e9794ad81542d71e83fdd1eccd143742ee3ac776ae087ab6deb73f794505aeL19). :slightly_smiling_face:

jpopelka: Oh, I see, it was there twice, and [one is still left](https://github.com/packit/dashboard/blob/c7cd0f33032e77754ed40e9106950413acd95bcd/files/ansible/install-deps.yaml#L15) :-)

I have actually one more question: do we still need the [npm](https://github.com/packit/dashboard/blob/c7cd0f33032e77754ed40e9106950413acd95bcd/files/ansible/install-deps.yaml#L14) when [yarn looks like a complete replacement](https://classic.yarnpkg.com/lang/en/docs/migrating-from-npm/)?

PR #184, opened 2022-06-22
Fix the server name in prod

csomh:
The server name in prod has no instance slug.

Signed-off-by: Hunor Csomortáni <csomh@redhat.com>


lachmanfrantisek (approved):
Thanks for the fix!

PR #191, opened 2022-09-29
Display multiple Copr builds for TF results if present

lbarcziova:
In packit/packit-service#1658 there was implemented that a TF run can be triggered with multiple Copr builds and API for TF runs changed (copr_build_id -> copr_build_ids) as well. This PR implements displaying those builds correctly in the TF result view.

Related to packit/packit-service#1658
 
The view for the TF runs with only one build stays the same and this is how it looks like for TF runs with more builds:
Snímka obrazovky z 2022-09-29 14-03-18

---

RELEASE NOTES BEGIN
N/A
RELEASE NOTES END

### DIFF
--- a/frontend/src/components/results/testing_farm.js
+++ b/frontend/src/components/results/testing_farm.js
@@ -68,20 +68,35 @@ const ResultsPageTestingFarm = (props) => {
         <>{data.status}</>
     );
 
-    const coprBuildInfo = data.copr_build_id ? (
-        <tr>
-            <td>
-                <strong>COPR build</strong>
-            </td>
-            <td>
-                <a href={`/results/copr-builds/${data.copr_build_id}`}>
-                    Details
-                </a>
-            </td>
-        </tr>
-    ) : (
-        ""
-    );
+    function getCoprBuilds() {
+        let coprBuilds = [];
+        let title =
+            data.copr_build_ids.length == 1 ? "Copr build" : "Copr builds";
+
+        for (var i = 0; i < data.copr_build_ids.length; i++) {
+            var coprBuildId = data.copr_build_ids[i];
+            var linkText =
+                data.copr_build_ids.length == 1
+                    ? "Details"
+                    : `Build ${i + 1}  `;
+            coprBuilds.push(
+                <Label href={`/results/copr-builds/${coprBuildId}`}>
+                    {linkText}
+                </Label>
+            );
+        }
+
+        return (
+            <tr>
+                <td>
+                    <strong>{title}</strong>
+                </td>
+                <td>{coprBuilds}</td>
+            </tr>
+        );
+    }
+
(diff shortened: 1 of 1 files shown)


frontend/src/components/results/testing_farm.js
```diff
@@ -68,20 +68,37 @@ const ResultsPageTestingFarm = (props) => {
...
+            data.copr_build_ids.length == 1 ? "Copr build" : "Copr builds";
+
+        for (var i = 0; i < data.copr_build_ids.length; i++) {
+            var coprBuildId = data.copr_build_ids[i];
+            var linkText =
+                data.copr_build_ids.length == 1
+                    ? "Details"
+                    : `Build ${i + 1}  `;
+            const coprBuildLink = (
+                <a href={`/results/copr-builds/${coprBuildId}`}>
```
mfocko, line 83: Could you probably wrap it in label, so that it looks a bit nicer? :) otherwise fine with me :)

mfocko (approved):

mfocko (approved):

lachmanfrantisek:
I know I am a bit late to the party, but can't we add more info to the labels?
(I really like that you've changed the visual style from the pure text.)

What about something like this?
Screenshot from 2022-09-30 15-01-47

Or maybe even with the numbers to be explicit?
Screenshot from 2022-09-30 15-02-48

Or use this syntax just with the extra build but that could complicate things.

@mfocko Do you know what is the difference between this and pipelines that we've lost the space between the labels?

lbarcziova:
@lachmanfrantisek the reason why I didn't want to do it that way was that it would require an additional API call since in the response for TF model we get only Copr build IDs. I was thinking about displaying the IDs instead of Build 1, 2, but on the other hand, the Copr build IDs are IDs in our DB which also does not make that much sense to show. Do you think it is worth it to do additional API calls/ changing the API so that we can display the PR number for the additional build (since you can see more info about the build right after clicking on the link for the build)?

lachmanfrantisek:
I am not sure if it is worth that extra call... Personally, it sounds logical to know what are these builds about and which one is the core one and which one is the extra one (what we don't know here, do we?). But we can leave it as is and see if anyone will ask for it...

TomasTomecek:
Can we enhance the existing API call so it includes the additional info and we wouldn't need to perform extra API calls?

mfocko:
why does the page load for so long /o\ :peek:

lbarcziova:
> Can we enhance the existing API call so it includes the additional info and we wouldn't need to perform extra API calls?

It is an option (as I already wrote in the previous comment -  changing the API), but the additional processing would be still there, just directly in the packit-service code and I'm not sure I like that kind of API design. I would wait for some feedback and we can revisit/even discuss on arch.

PR #203, opened 2023-01-14
chore: update JS dependencies to React 18

### LINKED ISSUE #202: NodeJS 18 not supported
We get obscure error if NodeJS 18 is used when trying to run or build the React server: `Error: error:0308010C:digital envelope routines::unsupported`

This is due to `babel-loader` that's being used has a version where an older and unsecure SSL plugin is being used. To fix this the dependencies used should be updated

SpyTec:
When using NodeJS 17+ we could not run development mode of the JS app as
it was using insecure SSL type. This changes that by updating the
dependencies to be React 18

The `yarn.lock` file has also been added, this is necessary to keep dependencies consistent between installations and machines. It should always be present if one is using `yarn` as package manager. If there is a reason it was added to gitignore due to infra reasons then infra is wrong

TODO:

- [ ] ‹fill in›

Fixes #202

Related to

Merge before/after

---

RELEASE NOTES BEGIN
Update React to v18 and related packages
RELEASE NOTES END


frontend/src/components/app_layout.js
```diff
@@ -50,9 +50,11 @@ const AppLayout = ({ children }) => {
                                 id={`${route.label}-${idx}`}
                             >
                                 <NavLink
-                                    exact
+                                    exact="true"
```
SpyTec, line 53: Newer React router complains that it cannot be a bool, and has to be a string

frontend/package.json
```diff
@@ -5,19 +5,19 @@
...
-    "react-dom": "^17.0.2",
-    "react-router-dom": "^5.2.0",
-    "react-scripts": "4.0.3",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
+    "react-router-dom": "^6.6.2",
+    "react-scripts": "5.0.1",
     "resize-observer-polyfill": "^1.5.1",
-    "web-vitals": "^1.0.1"
+    "web-vitals": "^3.1.1"
```
SpyTec, line 20: Have not tested to see if this works, not used it before. LMK what to do to make sure this works

frontend/package.json
```diff
@@ -5,19 +5,19 @@
   "dependencies": {
     "@patternfly/react-core": "^4.147.0",
     "@patternfly/react-icons": "^4.11.8",
```
SpyTec, line 7: I'll do a subsequent PR after this to update Patternfly

frontend/src/routes.js
```diff
@@ -48,61 +48,66 @@ const routes = [
 ];
 
 const AppRoutes = () => (
-    <Switch>
-        {routes.map(({ path, exact, component, title }, idx) => (
+    <Routes>
+        {routes.map(({ path, exact, element, title }, idx) => (
```
SpyTec, line 52: https://reactrouter.com/en/main/upgrading/v5#upgrade-all-switch-elements-to-routes

frontend/src/routes.js
```diff
@@ -18,28 +18,28 @@ import Pipelines from "./components/pipelines";
 // Main Menu routes
 const routes = [
     {
-        component: Dashboard,
+        element: <Dashboard />,
```
SpyTec, line 21: One thing of note here, we might need to change this to lazy load. I've not looked to see whether it is necessary or not

frontend/src/index.js
```diff
@@ -1,13 +1,13 @@
 import React from "react";
-import ReactDOM from "react-dom";
+import { createRoot } from "react-dom/client";
 import App from "./app";
 import reportWebVitals from "./reportWebVitals";
 
-ReactDOM.render(
+const root = createRoot(document.getElementById("root"));
+root.render(
     <React.StrictMode>
```
SpyTec, line 8: https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#updates-to-strict-mode

frontend/src/index.js
```diff
@@ -1,13 +1,13 @@
 import React from "react";
-import ReactDOM from "react-dom";
+import { createRoot } from "react-dom/client";
 import App from "./app";
 import reportWebVitals from "./reportWebVitals";
 
-ReactDOM.render(
+const root = createRoot(document.getElementById("root"));
+root.render(
```
SpyTec, line 7: The way they render has changed
https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#updates-to-client-rendering-apis

Makefile
```diff
@@ -5,14 +5,14 @@ TEST_TARGET ?= ./tests/
...
 	sudo dnf -y install python3-flask yarnpkg npm
 	cd frontend && yarn install
 	make transpile-prod
 
-install-logos:
-	ln -s $(CURDIR)/files/logos/stg/* $(CURDIR)/frontend/public/
-	ln -s $(CURDIR)/files/logos/stg.png $(CURDIR)/frontend/src/static/logo.png
+.install-logos:
+	ln -s -f $(CURDIR)/files/logos/stg/* $(CURDIR)/frontend/public/
+	ln -s -f $(CURDIR)/files/logos/stg.png $(CURDIR)/frontend/src/static/logo.png
```
SpyTec, line 15: If something goes wrong when doing `install-dependencies` after this has run, it will try to link again if you run it once more since the file exists. Force is necessary

Touch has also been added so that `.install-logos` won't execute if the file .install-logos exist as it's unnecessary. To run this again just delete the `.install-logos` file

SpyTec:
Also I have stuff I want to investigate and fix myself when I get the time, so having this merged soon would be great so I can start working without basing everything off of this branch

Makefile
```diff
@@ -5,14 +5,15 @@ TEST_TARGET ?= ./tests/
...
+	touch $@
```
SpyTec, line 16: If something goes wrong when doing `install-dependencies` after this has run, it will try to link again if you run it once more since the file exists. Force is necessary

Touch has also been added so that `.install-logos` won't execute if the file .install-logos exist as it's unnecessary. To run this again just delete the `.install-logos` file

jpopelka (approved):
Thank you! LGTM, but I'd like to ask @mfocko for a final word.

SpyTec:
> I wonder if we should keep both package.json and yarn.lock, or decide for one package manager and roll with it

So, package.json is the default no matter what you're using. It's not a lock file really. If you are using `npm` there will be a separate `package-lock.json` file, whereas with yarn there will be a `yarn.lock` file

TomasTomecek:
glad to hear package.json is the standard, didn't know that

TomasTomecek (approved):

frontend/src/components/app_layout.js
mfocko: ugh… but I guess it makes sense

frontend/src/routes.js
mfocko: IIRC it's just grid with static content and some links, not sure if it requires to be loaded lazily

mfocko (approved):
LGTM, thanks for the contribution :)

PR #207, opened 2023-01-20
refactor: fix eslint issues

SpyTec:
The linter had a lot of warnings when starting up the dev server. This
fixes all those warnings and makes sure we do things correctly

  - Array.map() was changed to Array.forEach() as it was only iterating and
not changing
  - Initial render call had explicit ignore added to clarify that it's a
  one-time render thing
  - Unused imports and functions removed
  - Incorrect usage of useState was fixed for columns (never needed
    state management for it)
  - checkValidProject was changed to be a callback and the useState
    using it was changed to correctly use it as a dependency

TODO:

- [ ] ‹fill in›

Note that #205 fixes some things for me, so until that is merged I cannot properly verify that the changes work as it should

Fixes

Related to

Merge before/after

---

RELEASE NOTES BEGIN
Fix ESLint warnings
RELEASE NOTES END

### DIFF
--- a/frontend/src/components/projects_list.js
+++ b/frontend/src/components/projects_list.js
@@ -30,7 +30,7 @@ const ProjectsList = (props) => {
     const [page, setPage] = useState(1);
     const [projects, setProjects] = useState([]);
     // by default, ignore non useful projects ie without any handled item
-    const [showUseful, setShowUseful] = useState(true);
+    const showUseful = true;
 
     // If a namespace and forge are provided, then load those
     // otherwise load all projects
@@ -84,6 +84,7 @@ const ProjectsList = (props) => {
     // here useEffect gets executed after page changes and it fetched more pages
     useEffect(() => {
         fetchData();
+        // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [page]);
 
     function loadMore() {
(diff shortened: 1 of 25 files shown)


jpopelka:
I don't see the described changes in the diff. What am I missing?

SpyTec:
> I don't see the described changes in the diff. What am I missing?

Seems like I didn't include it in the commit, fixed :grimacing:

frontend/src/components/projects_list.js
```diff
@@ -84,7 +84,8 @@ const ProjectsList = (props) => {
     // here useEffect gets executed after page changes and it fetched more pages
     useEffect(() => {
         fetchData();
-    }, [page]);
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, []);
```
SpyTec, line 88: If I'm thinking about the component correctly, this might not work. Noticed that there is some more stuff here I have to fix on top of this

SpyTec: I just reverted this for now. Would need to rewrite a lot and this is good enough

jpopelka (approved):
:+1: Thanks!

add `mergeit` label once you're ready

SpyTec:
@jpopelka I need to be a member to add labels

PR #218, opened 2023-02-17
feat: accessibility improvements

### LINKED ISSUE #217: Label groups in tables are not accessiblity friendly
Within the labels we have groups, such as `Testing Farm`, and within that we have labels that are green or red depending if they passed or not. There is one issue where it seems like screen readers can't read the status of the labels, like `fedora-36-x86_64` saying success, so they have no idea what the current status of it is

What could be used is this, which we can show within each label. E.g. `fedora-36-x86_64 - success`. Where the success part isn't shown visually

https://www.patternfly.org/v4/utilities/accessibility/#screen-reader-only

SpyTec:
This adds more information to the dashboard for screen readers. It also changes the description of the various tables as the screen readers already read out the word table

Fixes #217

Related to

Merge before/after

---

RELEASE NOTES BEGIN
Improve accessibility for tables and its content
RELEASE NOTES END

### DIFF
--- a/frontend/src/components/dashboard.js
+++ b/frontend/src/components/dashboard.js
@@ -27,7 +27,7 @@ const Dashboard = () => (
         <PageSection>
             <Grid hasGutter>
                 <GridItem lg={3}>
-                    <Card>
+                    <Card isFullHeight>
                         <CardHeader>
                             <CatalogIcon />
                             &nbsp;<b>Documentation</b>
@@ -42,7 +42,7 @@ const Dashboard = () => (
                 </GridItem>
 
                 <GridItem lg={3}>
-                    <Card>
+                    <Card isFullHeight>
                         <CardHeader>
                             <BellIcon />
                             &nbsp;<b>Blog posts</b>
@@ -58,7 +58,7 @@ const Dashboard = () => (
                 </GridItem>
 
                 <GridItem lg={3}>
-                    <Card>
+                    <Card isFullHeight>
                         <CardHeader>
                             <BugIcon />
                             &nbsp;<b>Issues</b>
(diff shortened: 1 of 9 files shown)


frontend/src/components/dashboard.js
```diff
@@ -27,7 +27,7 @@ const Dashboard = () => (
         <PageSection>
             <Grid hasGutter>
                 <GridItem lg={3}>
-                    <Card>
+                    <Card isFullHeight>
```
SpyTec, line 30: This is actually not part of it but just makes it more uniform on the homepage

image

frontend/src/components/projects_list.js
```diff
@@ -126,16 +126,17 @@ const ProjectsList = (props) => {
             <Gallery hasGutter>
                 {projects.map((project, index) => (
                     <GalleryItem key={index}>
-                        <Card isHoverable>
+                        <Card>
                             <CardTitle>
-                                <Link to={() => goToProjectInfo(project)}>
+                                <Link to={goToProjectInfo(project)}>
```
SpyTec, line 131: It should be evaluated right away, otherwise the link doesn't work as it sends in a function to the href field which will never be executed

frontend/src/components/projects_list.js
```diff
@@ -146,6 +147,7 @@ const ProjectsList = (props) => {
                                         <Tooltip
                                             position={TooltipPosition.top}
                                             content={"Branches Handled"}
+                                            aria="labelledby"
```
SpyTec, line 150: This doesn't really work as links are not present. So it has nothing to label. For screen readers they will either not see anything at all or just see a number AFAIK

frontend/src/components/tables/copr.js
```diff
@@ -134,7 +134,7 @@ const CoprBuildsTable = () => {
     return (
         <div>
             <Table
-                aria-label="Table of Copr builds"
+                aria-label="Copr builds"
```
SpyTec, line 137: Screen reader was reading out "Table Table of" when I tried, so it was overly verbose

jpopelka (approved):
Thank you! :+1:

mfocko (approved):

PR #219, opened 2023-03-01
Pull from upstream integration

### LINKED ISSUE #199: Show pull from upstream
Implement showing the new job in the dashboard similarly as propose downstream. The data are already stored in the packit-service DB, but there needs to be implemented a new API endpoint (the existing functionality for propose downstream should be only reused).

Make sure:
- [ ] there is a separate view in the [/jobs](https://dashboard.packit.dev/jobs) page 
- [ ] the job is shown in the [/pipelines](https://dashboard.packit.dev/pipelines) page
- [ ] there is a page with details about a job run (accessible from both pipelines and jobs view)

---
*Part of packit/dashboard#187 epic.*

lbarcziova:
Create downstream jobs menu item and route (dashboard.packit.dev/downstream-jobs) and add a new pull from upstream table to it. Modify the propose downstream table and result view so that they can be reused for pull from upstream. Also show the job in the /pipelines view properly.

Fixes #199

Snímka obrazovky z 2023-03-01 11-10-04

---

RELEASE NOTES BEGIN
TBD
RELEASE NOTES END


lbarcziova:
Since the `/jobs` already have multiple tabs and to me, it makes sense to separate the upstream and downstream jobs, I created the `/downstream-jobs` and added the pull from upstream results there (and this would be the place to add the koji builds and bodhi updates). WDYT, can it stay separate? And what about the naming? I didn't want to change the `/jobs` but it would maybe make sense to have `Upstream jobs` and `Downstream jobs` :thinking:

SpyTec:
> Since the `/jobs` already have multiple tabs and to me, it makes sense to separate the upstream and downstream jobs, I created the `/downstream-jobs` and added the pull from upstream results there (and this would be the place to add the koji builds and bodhi updates). WDYT, can it stay separate? And what about the naming? I didn't want to change the path `/jobs` but it would maybe make sense to change the title at least and have `Upstream jobs` and `Downstream jobs` 🤔

I refactored `/jobs` a bit in #224 to make life easier to code in it and its UX, so might be worth looking a bit at that PR and how it's setup as well

Anyhow, I think that we could probably use the [grouped or expandable functionality](https://www.patternfly.org/v4/components/navigation#grouped) of the navigation in the sidebar to separate upstream and downstream without making it all confusing. Either way the `/jobs` URL needs to remain the same to preserve backwards compatibility (or at least have redirects setup)

As for the tabs, they should be omitted until we get more tabs tbh

image

lbarcziova:
The grouped functionality really looks like a good fit :+1:  The grouping could be also done the other way, e.g.
```
Jobs
   Upstream
   Downstream
```


I think it would be worth discussing in our architecture meeting on Thu. Therefore, also looking into into visual changes of the "tabs" in #224, after merging that PR I will probably add pull-from-upstream into the existing `Jobs` view for now.

PR #213, opened 2023-02-03
feat: load indicator using React Query

SpyTec:
This starts using React Query for utilizing requests to the Packit API.
If the user is on a page and alt tabs currently or is inactive for a
while the page will refresh the content automatically as well. This can
be disabled either for each query individually or modified globally.

With React Query we also use caching to a better extent if it is ever
needed.

When the demo loads pages 1 through 3 it is due to alt tabbing to showcase the refresh of pages

TODO:

- [x] Other tabs in jobs page
- [x] Approval on introducing React Query dependency

https://user-images.githubusercontent.com/6598829/216692107-8fb0c5ca-000c-41f2-8ee9-84a052a65dd8.mp4

Fixes #212 #211 #210

Related to

Merge before/after

---

RELEASE NOTES BEGIN
Improve fetching of API content in paginated tables
RELEASE NOTES END

PR #220, opened 2023-03-01
Usage page

### LINKED ISSUE #193: Create a subpage with basic info about Packit usage
Part of https://github.com/packit/dashboard/issues/192

* [ ] Create a new subpage to show a basic about Packit usage/adoption
* [ ] Use data from the newly created API endpoint implemented in packit/packit-service#1724

lachmanfrantisek:
Fixes #193 

I would be really glad if someone more knowledgeable about React (=any experience is better than mine..;) can take a look and patiently review the code..;)

For the purpose of caching and easier work with dates, I am proxying the API calls through the backend which results in an issue with CORS. I hope using a flask-cors extension for that is fine.

Sadly, monthly and yearly data can't be accessed from prod for now..;(

---

RELEASE NOTES BEGIN
There is a new subpage on our dashboard you can visit to check how much is Packit used.
RELEASE NOTES END

---

Here is a screenshot with the data from prod instance:
image

### DIFF
--- a/frontend/src/components/tables/usage.js
+++ b/frontend/src/components/tables/usage.js
@@ -25,14 +25,14 @@ import { StatusLabel } from "../status_labels";
 import { Timestamp } from "../../utils/time";
 import { useParams } from "react-router-dom";
 
-const UsageComponent = () => {
+const UsageComponent = (props) => {
     const [hasError, setErrors] = useState(false);
     const [loaded, setLoaded] = useState(false);
     const [data, setData] = useState({});
     const top = 5;
 
     useEffect(() => {
-        fetch(`${process.env.REACT_APP_API_URL}/usage?top=${top}`)
+        fetch("http://127.0.0.1:5000/api/usage/" + props.what)
             .then((response) => response.json())
             .then((data) => {
                 setData(data);
--- a/frontend/src/components/tables/usage.js
+++ b/frontend/src/components/tables/usage.js
@@ -16,6 +16,7 @@ import {
     TableVariant,
     cellWidth,
 } from "@patternfly/react-table";
+import { ChartDonut } from "@patternfly/react-charts";
 
 import ConnectionError from "../error";
 import Preloader from "../preloader";
@@ -24,19 +25,7 @@ import { StatusLabel } from "../status_labels";
 import { Timestamp } from "../../utils/time";
 import { useParams } from "react-router-dom";
 
-const UsageTable = () => {
-    // Headings
-    const columns = [
-        {
-            title: <span className="pf-u-screen-reader">Forge</span>,
-            transforms: [cellWidth(5)],
-        }, // space for forge icon
-        { title: "Trigger", transforms: [cellWidth(15)] },
-        { title: "Chroots", transforms: [cellWidth(60)] },
-        { title: "Time Submitted", transforms: [cellWidth(10)] },
-        { title: "Copr Build", transforms: [cellWidth(10)] },
-    ];
-
+const UsageComponent = () => {
     const [hasError, setErrors] = useState(false);
     const [loaded, setLoaded] = useState(false);
     const [data, setData] = useState({});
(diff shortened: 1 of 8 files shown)


frontend/src/components/tables/usage.js
```diff
@@ -0,0 +1,227 @@
...
+        fetch("http://127.0.0.1:5000/api/usage/" + props.what)
+            .then((response) => response.json())
+            .then((data) => {
+                setData(data);
+                setLoaded(true);
+            })
+            .catch((err) => {
+                console.log(err);
+                setErrors(err);
+            });
```
SpyTec, line 44: Similar to my PR of #213 it should simplify data fetching within this component if that is used too, takes away a lot of the complexity

frontend/src/components/tables/usage.js
```diff
@@ -0,0 +1,227 @@
...
+        <div>
+            <Card>
+                <CardBody>
+                    {all_projects_instance_chart}
+                    {active_projects_instance_chart}
+                </CardBody>
+            </Card>
+            <Card>
+                <CardBody>{job_charts}</CardBody>
+            </Card>
```
SpyTec, line 222: There should be a way to get it more responsive in how it displays everything by having each chart in its own card. I can try to see how it would work next week

frontend/src/components/tables/usage.js
```diff
@@ -0,0 +1,227 @@
...
+                .replace(" Targets", "s")
+                .replace("Vm", "VM")
+                .replace("Tft", "TFT")
+                .replace("Srpm", "SRPM"),
+            data.jobs[key].job_runs
+        )
+    );
+
+    return (
+        <div>
```
SpyTec, line 213: Since the `div` here is just for wrapping multiple nodes, there is actually a React Element that can render the elements without needing to render the wrapping node. Can use either `<React.Fragment>` or `<>`

https://reactjs.org/docs/fragments.html

lachmanfrantisek: Thanks, good to know!

frontend/src/components/tables/usage.js
lachmanfrantisek: Thanks a lot! I've tried my best to mimic that... but since my usaces doesn't use paging, I am really not sure what I am doing here..;)

frontend/src/components/tables/usage.js
lachmanfrantisek: Yes, I agree. Especially since the project names can be really long.

I would be really glad if you have any suggestions... I've tried to play with the single cards for charts but wasn't much successful.

@SpyTec Thanks a lot in general for helping us with the dashboard!

frontend/src/components/tables/usage.js
```diff
@@ -0,0 +1,200 @@
...
+            (response) => response.json(),
+        );
+
+    const { data, isLoading, isError } = useQuery(
+        "usage" + props.what,
+        fetchData,
+        {
+            keepPreviousData: true,
+        },
+    );
```
SpyTec, line 20: Looking good! I would change it up a bit so that the `fetchData` isn't tied to the component, which also helps with performance. Overall changing the `what` prop to granularity might also be a good idea, as `what` doesn't tell much.
```suggestion
const fetchDataByGranularity = (granularity) =>
    fetch("http://127.0.0.1:5000/api/usage/" + granularity).then(
        (response) => response.json(),
    );
const UsageComponent = (props) => {
    const { data, isLoading, isError } = useQuery(
        "usage" + props.what,
        () => fetchData(props.what),
        {
            keepPreviousData: true,
        },
    );
```

frontend/src/components/tables/usage.js
lachmanfrantisek: OK, I've tried using `Flex` and it's a bit better.

frontend/src/components/tables/usage.js
lachmanfrantisek: Thanks!

lachmanfrantisek:
image
image

lachmanfrantisek:
And these are using prod data if there is anyone interested (looks like we are now able to load all the data :rocket: :tada: ):
image
image
image
image

lbarcziova (commented):
The charts look great, nice job!! Can't review the code to details, but to me, it looks good :+1:

Just thinking out loud, as we recently changed the tabs to be in separate, linkable, pages (3e6d8113daff006ed66e7195161fa1cdfe20852b), do you think this could be also useful for this use-case or linking of the time-specific usage charts would not be that beneficial?

lachmanfrantisek:
> Just thinking out loud, as we recently changed the tabs to be in separate, linkable, pages ([3e6d811](https://github.com/packit/dashboard/commit/3e6d8113daff006ed66e7195161fa1cdfe20852b)), do you think this could be also useful for this use-case or linking of the time-specific usage charts would not be that beneficial?

Good point... I am not sure. On one hand, we can/should be consistent, on the other hand, I don't expect someone needs to link separate tabs. I consider this as the first version and I expect changes to the structure of the page... With that, if no one has a strong opinion I would probably leave it as is for now...

frontend/src/components/tables/usage.js
```diff
@@ -0,0 +1,219 @@
...
+                .replace(" Targets", "s")
+                .replace("Vm", "VM")
+                .replace("Tft", "TFT")
+                .replace("Srpm", "SRPM"),
+            data.jobs[key].job_runs,
+        ),
+    );
+
+    return (
+        <>
```
TomasTomecek, line 201: just curious, what is this? :D

SpyTec: [React Fragments!](https://reactjs.org/docs/fragments.html) A way to wrap elements in React without creating a DOM Element for the user

packit_dashboard/api/routes.py
```diff
@@ -61,3 +68,46 @@ def propose_downstream():
...
+    ).strftime("%Y-%m-%d")
+    return _get_usage_data_from_packit_api(usage_from=last_month_date)
+
+
+@api.route("/api/usage/last-year")
+@ttl_cache(maxsize=_CACHE_MAXSIZE, ttl=timedelta(days=1).seconds)
+def usage_last_year():
+    now = datetime.now()
+    last_year_date = now.replace(year=now.year - 1).strftime("%Y-%m-%d")
+    return _get_usage_data_from_packit_api(usage_from=last_year_date)
```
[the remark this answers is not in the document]
SpyTec: Agreed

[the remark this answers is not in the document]
lachmanfrantisek: Ok, I'll rename all of the intervals, then...

past day / past week / past month / past year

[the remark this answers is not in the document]
lachmanfrantisek: Fixed.

TomasTomecek (approved):
:rocket: :chart_with_upwards_trend: 

Thank you!

frontend/src/components/tables/usage.js
```diff
@@ -0,0 +1,219 @@
...
+import { ChartDonut } from "@patternfly/react-charts";
+
+import ConnectionError from "../error";
+import Preloader from "../preloader";
+import { useQuery } from "react-query";
+
+const fetchDataByGranularity = (granularity) =>
+    fetch("http://127.0.0.1:5000/api/usage/" + granularity).then((response) =>
+        response.json(),
+    );
```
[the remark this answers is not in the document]
lachmanfrantisek: Yes..;) Just found out that it looks for my local backend...;)

Using `${process.env.REACT_APP_API_URL}` should do the trick, right?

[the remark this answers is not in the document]
SpyTec: Yup, should work fine!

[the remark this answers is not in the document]
lachmanfrantisek: Thanks!

Sadly, this returns Packit API, not the API of the backend... I can use just `/api/usage/${granularity}` which should work well when deployed, but I can't make it work locally because of a different port for frontend/backend...;(

PR #224, opened 2023-03-03
Refactor React Router usage

### LINKED ISSUE #221: React Router links are inconsistently used
Some label links etc. that link to internal resources cause a refresh rather than an in-page reload that React Router can allow

Likewise, the Brand logo in the navigation bar also doesn't link with React Router, thus causing a page reload

### LINKED ISSUE #222: Sidebar does not show currently active page
When going to Jobs or Pipelines, the sidebar doesn't indicate that you are on the correct page. E.g. when applying the isActive to the NavItem it should show like this

image

### LINKED ISSUE #223: Refactor Jobs page to allow for linking to specific tables like Copr or Koji
This would speed up having to go to Koji if someone links to the Jobs page itself. This could be done by having a URL like `/jobs/copr-builds`

SpyTec:
This fixes a bunch of different things. Specifically the issues I opened today

Most of the links within the site now default to using React Router Link or NavLink where appropriate, preventing any page reload. Fixes #221

There is now a way to see current page on the sidebar if you are on pipelines or jobs page for instance. Fixes #222

Jobs tables have been updated to allow linking to specific tables. I also took the liberty of refactoring the website route structure to now fully utilize Outlet instead of rendering components wherever needed and having to deal with other issues like the tables would have been. Fixes #223

TODO:

- [ ] Going to a page that is not in the sidebar will not make sidebar change the style of isActive, meaning that if you go to a results page it will still show that the last page is active in the sidebar

Fixes

Related to

Merge before/after

---

RELEASE NOTES BEGIN
Refactor React Router to use Outlet for routes
Add links for job tables
RELEASE NOTES END

### DIFF
--- a/frontend/src/components/app_layout.js
+++ b/frontend/src/components/app_layout.js
@@ -1,5 +1,4 @@
-import * as React from "react";
-import { NavLink } from "react-router-dom";
+import { NavLink, Outlet, matchRoutes, useLocation } from "react-router-dom";
 import {
     Brand,
     Nav,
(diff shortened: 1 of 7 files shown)


frontend/src/components/app_layout.js
```diff
@@ -12,14 +11,33 @@ import {
...
+    useEffect(() => {
+        // it matches everything that has to do with the current URL. Start from the leaf and work up to set title if at all.
+        for (let index = currentRouteTree.length - 1; index >= 0; index--) {
+            const element = currentRouteTree[index];
+            if (element.route.label) {
+                setActiveLocationLabel(element.route.label);
+            }
+            if (element.route?.title) {
+                document.title = element.route.title;
+                break;
```
SpyTec, line 30: Title is like this

image

SpyTec:
Once this is merged in some capacity I'll spin up a PR to handle errors too using [`errorElement`](https://reactrouter.com/en/main/route/error-element)

lbarcziova (approved):
looks great, I really like that we will be able to link specific tables, thanks!

Can we merge this already so that I can use the same in #219 or is anything else needed here?

SpyTec:
> looks great, I really like that we will be able to link specific tables, thanks!
> 
> Can we merge this already so that I can use the same in #219 or is anything else needed here?

Unless the TODO is a blocker in your eyes I would say it's good to merge! The TODO is just about the sidebar isActive. Can fix that in a follow-up PR

PR #233, opened 2023-03-08
Use correct backend URL when fetching usage data

lachmanfrantisek:
This is a fix to really access backend when getting data from backend's API (=flask).


mfocko (approved):
LGTM

PR #235, opened 2023-03-09
feat: Catch app crashes

### LINKED ISSUE #234: Show error page on app crashes
This should be something a client should never see, but the unexpected might happen. With #224 we can now use `errorElement` and show a component if the app crashes

SpyTec:
If the app crashes we casn now have a error element that can display to the user. This will display a log of what happened, and also let the user know they can report it upstream and also go back to the main page.

First commit shows the feature
Second commit shows refactor of existing connection error 

App crash exception
image

Refactor of connection issue error
image

TODO:

- [ ] ‹fill in›

Fixes #234 

Related to

Merge before/after

---

RELEASE NOTES BEGIN
Add user friendly interface if dashboard crashes
RELEASE NOTES END

### DIFF
--- a/frontend/src/components/branch.js
+++ b/frontend/src/components/branch.js
@@ -1,6 +1,6 @@
 import React, { useState, useEffect } from "react";
 
-import ConnectionError from "./error";
+import ConnectionError from "./error_connection";
 import Preloader from "./preloader";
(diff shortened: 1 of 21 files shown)


lachmanfrantisek (approved):
Thank you!

PR #240, opened 2023-03-24
feat(Storybook): Initial setup of how storybook can be used

SpyTec:
Basic concept of what storybooks are. I've not added a makefile yet so to run go to frontend/ and run `yarn storybook`

If you are in the root repository, you can do `make storybook`. I might change this to dev-storybook or something.

TODO:
- [x] Flesh out the stories a bit more
- [x] Put test data outside the stories components

Fixes #228 

Related to

Merge after
#246 

---

RELEASE NOTES BEGIN
Setup Storybook for component development and showcase
RELEASE NOTES END

PR #246, opened 2023-04-04
chore: change to Vite from CRA

SpyTec:
Create React App (CRA) is being deprecated and Vite now works fine with
PatternFly. This should make overall development faster and also allow
some more flexibility

Also just saw this, might be something to consider for the api instead of using env variables?
https://vitejs.dev/config/server-options.html#server-proxy

TODO:

- [x] Fix references to files in `/public` folder
- [x] Fix environment variable usage if you don't use Makefile
- [x] Remove dependencies from and around CRA

- https://vitejs.dev/
- https://blog.bitsrc.io/the-future-of-react-why-create-react-app-is-deprecated-and-hooks-are-the-future-83e8a087a325

Fixes

Related to

Merge before
#240 

Merge after
#242 

---

RELEASE NOTES BEGIN
Change dev and build environment from Webpack to Vite
RELEASE NOTES END


SpyTec:
@mfocko ready for Vite change!

jpopelka (approved):

PR #252, opened 2023-05-19
chore: upgrade React Query to v4

### LINKED ISSUE #243: Update React dependency of `react-query` from v3 to v4
Turns out that React Query changed their packaging from `react-query` to `@tanstack/react-query`. We need to use this to upgrade from v3 to v4. I don't think there will be any major advantages to upgrading looking at changelog, but should still be done

There is a codemod we can use to automate the tedious bits

https://tanstack.com/query/v4/docs/react/guides/migrating-to-react-query-4

SpyTec:
As part of React Query v4 they changed the package name itself from just `react-query` to `@tanstack/react-query`. We now use the new name and updated all existing queries to the new format.

TODO:

- [ ] ‹fill in›

Fixes
#243 

Related to

Merge before/after

---

RELEASE NOTES BEGIN
Upgrade React Query dependency from v3 to v4
RELEASE NOTES END


jpopelka (approved):
:+1: thanks!

PR #253, opened 2023-05-23
ci: add GH Action for deploying PRs

mfocko:


SpyTec:
Is the GitHub Pages enabled?

There is also this which is made by GitHub itself, might be better
https://github.com/marketplace/actions/branch-deploy

mfocko:
> Is the GitHub Pages enabled?
> 
> There is also this which is made by GitHub itself, might be better https://github.com/marketplace/actions/branch-deploy

Not yet, it doesn't matter; since the PR is from fork `GITHUB_TOKEN` will not get `write` permissions anyways :/

mfocko:
> Otherwise, we might be able to use surge as PatternFly does https://github.com/patternfly/patternfly/blob/v5/.github/workflows/pr-preview.yml https://github.com/patternfly/patternfly/blob/v5/.github/upload-preview.js

I was thinking about that one, might need to look into it, thanks a lot for the tip :)

PR #256, opened 2023-05-23
fix: propose downstream crashed the application

### LINKED ISSUE #254: Failed release syncs not displayed and causing crash
Example: https://dashboard.packit.dev/results/propose-downstream/2533
Reason should be: `TypeError: e.link is null`

SpyTec:
This happened when the URL was null and we tried to use `startsWith` on it. For this to be fixed we have to assert that it is not null and proceed with the assumption that it can be null.

Specifically, this happens when propose downstrean doesn't have the `downstream_pr_url` attribute set.

TODO:

- [ ] ‹fill in›

Fixes
#254

Related to

Merge before/after

---

RELEASE NOTES BEGIN
Fix propose downstream detailed view from crashing when downstream URL is null
RELEASE NOTES END

### DIFF
--- a/frontend/src/app/StatusLabel/BaseStatusLabel.tsx
+++ b/frontend/src/app/StatusLabel/BaseStatusLabel.tsx
@@ -3,7 +3,7 @@ import { Label, Tooltip } from "@patternfly/react-core";
 import { Link } from "react-router-dom";
 
 export interface BaseStatusLabelProps {
-    link: string;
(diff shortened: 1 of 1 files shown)


nforro (approved):
I don't know anything about React but the change looks reasonable to me :slightly_smiling_face:

lbarcziova (approved):
thanks!

PR #267, opened 2023-05-28
feat: change projects from cards to table

SpyTec:
We were using cards for the list itself, together with no differentiator
for each project, such as their icon, it made it very difficult to
browse in an easy way.

This change moves it over to a composable Table which we can use in
a future change to automatically fetch the branches, releases, and
issues for each project without leaving the page.

However, the table itself offers more information and gives less
confusion as to what each entry is and where to look to see the name --
the repository name will always be on the left side now, which it
didn't before

image

TODO:

- [ ] ‹fill in›

Fixes

Related to

Merge before/after

---

RELEASE NOTES BEGIN
Our dashboard now shows projects in a table instead of the cards.
RELEASE NOTES END

### DIFF
--- a/frontend/src/app/Projects/ProjectSearch.tsx
+++ b/frontend/src/app/Projects/ProjectSearch.tsx
@@ -9,6 +9,10 @@ import {
     GridItem,
     Button,
     TextInput,
+    Panel,
+    InputGroup,
+    PanelMain,
+    PanelMainBody,
 } from "@patternfly/react-core";
 
 import { SearchIcon } from "@patternfly/react-icons";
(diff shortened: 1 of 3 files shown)


jpopelka (approved):

mfocko:
@SpyTec I reverted an addition of one entry to the gitignore from 519a111. I think it was committed by mistake.

SpyTec:
> @SpyTec I reverted an addition of one entry to the gitignore from 519a111. I think it was committed by mistake.

https://github.com/mswjs/msw/discussions/1015

Here is the discussion I found so I would say leave it in. I just thought I added it with a comment before. My bad

mfocko:
got it, will add it back

PR #272, opened 2023-05-30
fix(gha): build args »need« to be on separate lines

mfocko:
In 67650a1, they have been put on one line, this doesn't work according to the documentation… and has already bitten me once before.

Revert this change and split them to multiple separate lines.


SpyTec (approved):

jpopelka (approved):
I'll be calling you Mato The Lightning :)

PR #273, opened 2023-06-05
Don't install nodejs:18 (RPM) module

### LINKED ISSUE #247: Can't base image on F38 due to missing nodejs:18 module
### Description

We're using [nodejs:18 module](https://github.com/packit/dashboard/blob/main/files/ansible/install-deps.yaml#L9).

Currently, there's no `nodejs:18` module in F38, only `nodejs:16`, see `podman run -ti --rm fedora:38 dnf module list`.
Does anyone know where to request it?

I'm also wondering what's the difference between the `nodejs:18` module and `nodejs` rpm which also looks like version 18.
BTW, there's also `nodejs20` in both F37 and F38, can't we use that?

### Benefit

We can bump base image to F38

### Importance

_No response_

### Workaround

- [ ] There is an existing workaround that can be used until this feature is implemented.

### Participation

jpopelka:
F37 already contains `nodejs` 18 and I haven't seen any build problems.
C9S (#258) contains `nodejs` 16 but everything looks OK as well.

Fixes #247

Reverts 3cd34b74

@mfocko do you have any idea why @csomh started deleting the `/src/frontend/node_modules` in the 3cd34b74 and whether I should revert it too? I left it untouched (i.e. we still delete it) and it looks OK.

### DIFF
--- a/files/ansible/common.yaml
+++ b/files/ansible/common.yaml
@@ -15,7 +15,7 @@
   tags:
     - no-cache
   register: src_path
-- name: Let'secrets make sure {{ packit_dashboard_path }} is present
+- name: Make sure {{ packit_dashboard_path }} is present
   assert:
     that:
       - "src_path.stat.isdir"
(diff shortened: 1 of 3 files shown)


mfocko:
We don't need the `node_modules`, cause after react gets built, it's a static site.

mfocko (approved):

Issue #269, opened 2023-05-30
1. Link Usage page in the menu


lbarcziova:
Create a new item in the menu `Usage` which will link the `/usage`.

PR #293, opened 2023-06-17
Add label for Usage menu

c3d:
Add label for Usage menu so that it shows up

FIxes: #269


stefwalter:
Could you provide a screenshot? I can't find the Usage menu item.

SpyTec (commented):
LGTM :tada:

SpyTec (approved):

c3d:
> Could you provide a screenshot? I can't find the Usage menu item.

The link is there, but it does not seem to be doing what we expect.
PR 269

SpyTec:
> > Could you provide a screenshot? I can't find the Usage menu item.
> 
> The link is there, but it does not seem to be doing what we expect. PR 269

It's a bit funky since it fetches from backend. We'd need to use another service. I'll merge it now and when it shows up on the big screen it should work :D

PR #298, opened 2023-06-18
Revert getXYZLink() to again accept forge instead of gitRepo

jpopelka:
I changed my mind after all :-)

Passing repo URL, namespace and name is redundant because namespace and name are in the URL. We could pass only the URL, but splitting it isn't that easy because the namespace can also contain the slash on GitLab.


jpopelka:
Closing in favor of #300

Issue #262, opened 2023-05-24
3. Display submitted time for Testing Farm results


lbarcziova:
Currently, in `/results/testing-farm/{id}` we do not display any time regarding the specific run (e.g. https://dashboard.packit.dev/results/testing-farm/216017).

This information is available via the API already (`submitted_time`), so let's show it also to be consistent with other result pages (for builds, release syncs).

PR #303, opened 2023-06-21
feat: Consistent results pages

### LINKED ISSUE #288: 9. Use the Card for SRPM result page
The result page looks different for Copr and SRPM:

image
image

Let's use the card on the SRPM result page as well.
* Copr result page: https://github.com/packit/dashboard/blob/b884ed00977eb18f4c27e982e89c25de01cc63a0/frontend/src/app/Results/ResultsPageCopr.tsx
* SRPM result page: https://github.com/packit/dashboard/blob/b884ed00977eb18f4c27e982e89c25de01cc63a0/frontend/src/app/Results/ResultsPageSRPM.tsx

SpyTec:
This unifies the results to a better UX that should make it less confusing
while also having a consistent layout between different builds and runs

This idea is a change from table to a [description list](https://www.patternfly.org/v4/components/description-list)
image

Before and after for testing farm
image

Before and after for Copr builds
image

TODO:

- [ ] ‹fill in›

Fixes 
#288
#262

Related to

Merge before/after

---

RELEASE NOTES BEGIN
Improve UX of results page
RELEASE NOTES END

### DIFF
--- a/frontend/src/app/Results/ResultsPageSyncReleaseRuns.tsx
+++ b/frontend/src/app/Results/ResultsPageSyncReleaseRuns.tsx
@@ -14,6 +14,10 @@ import {
     Button,
     ToolbarGroup,
     Tooltip,
+    DescriptionList,
+    DescriptionListDescription,
+    DescriptionListGroup,
+    DescriptionListTerm,
 } from "@patternfly/react-core";
 
 import { ErrorConnection } from "../Errors/ErrorConnection";
@@ -102,7 +106,9 @@ const ResultsPageSyncReleaseRuns: React.FC<ResultsPageSyncReleaseRunsProps> = ({
     }
 
     const linkToDownstreamPR = data.downstream_pr_url ? (
-        <a href={data.downstream_pr_url}>Click here</a>
+        <a href={data.downstream_pr_url} rel="noreferrer" target="_blank">
+            Click here
+        </a>
     ) : (
         <>Link will be available after successful downstream PR submission.</>
     );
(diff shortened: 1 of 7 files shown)


frontend/src/app/Results/ResultsPageSyncReleaseRuns.tsx
```diff
@@ -145,65 +153,63 @@ const ResultsPageSyncReleaseRuns: React.FC<ResultsPageSyncReleaseRunsProps> = ({
...
-                        </table>
+                            <DescriptionListGroup>
+                                <DescriptionListTerm>
+                                    Status
+                                </DescriptionListTerm>
+                                <DescriptionListDescription>
+                                    {data.status}
+                                </DescriptionListDescription>
+                            </DescriptionListGroup>
+                            <DescriptionListGroup>
```
lachmanfrantisek, line 170: image


Looks like the formatting somehow split the items.

lachmanfrantisek (commented):
Thanks!

Do I remember correctly that we've used this approach before?..;)

Personally, I don't have a strong opinion on this.

The thing that I am not sure about in general is how to call the text for links.
E.g. with the following example, we are duplicating the words in the header and the value:

. | .
-- | --
Copr URL| [Web URL](https://copr.fedorainfracloud.org/coprs/build/6103664/)
Build Logs | [Build Logs URL](https://download.copr.fedorainfracloud.org/results/packit/packit-hello-world-1516/fedora-rawhide-x86_64/06103664-hello/builder-live.log)


What about using  this:

. | .
-- | --
Copr | `Web URL` `Build Logs`

(E.g. by using labels.)

SpyTec:
What about something like 

**Builds**

Copr (logs)

lachmanfrantisek:
> **Builds**
> 
> Copr (logs)

Yes, why not. I quite like the labels but this will work fine as well. I hope people will realise that these are two separate links.

SpyTec:
I will check with PatternFly or UX designers at some point to find best course of action

------- Original Message -------

SpyTec:
@lachmanfrantisek WDYT

mfocko:
> Do I remember correctly that we've used this approach before?..;)

I tried, IIRC you were against :D

SpyTec:
@mfocko for SRPM builds, I noticed a lot of the fields can be "Not available". Pretty much all of them tbh

Is this when the build isn't ready yet? If so, can we determine that by a specific field? Right now the code checks every variable which isn't ideal as it makes it seem like everything is conditional and can or cannot be available at random

mfocko:
I guess the SRPM depends on the finishing time, but that build can still fail and there wouldn't be any SRPM in that case

SpyTec:
Okay, I mean if there are a lot of edge-cases I will double-check that as I fix (or just keep the logic the same)

 I updated the Testing Farm layout so it shows this now, allowing you to fetch builds in-line, WDYT? I want to do this on more things

https://github.com/packit/dashboard/assets/6598829/227aead4-a242-4e25-b0cc-5e8a8bda4e57

lachmanfrantisek:
I really like the simplifications you are doing!

* I would move the commit sha away from the table. (@majamassarini is also working on a similar change on the database level since the commit is not strictly related to the result.)
* I like the Copr build visualisation for multiple builds, but we should put something interesting into the detail so it's useful to click on it. And also, make it look good for one/no build since that's the most used option.
* Not sure if someone would copy-paste the id of the test but I like that.

SpyTec:
>     * I would move the commit sha away from the table. (@majamassarini is also working on a similar change on the database level since the commit is not strictly related to the result.)

Do we have an alternative?

>     * I like the Copr build visualisation for multiple builds, but we should put something interesting into the detail so it's useful to click on it. And also, make it look good for one/no build since that's the most used option.

For sure! 

>     * Not sure if someone would copy-paste the id of the test but I like that.

There are some use-cases for it, @danmyway mentioned to me that it would be useful for CLI stuff

danmyway:
I agree with @SpyTec.
We have a CLI Testing Farm result parser, so we could use the copy-paste field as a source.

lachmanfrantisek:
> Do we have an alternative?

Maybe something like the following to follow GitHub style of showing commits:
image

(But I am not sure how  we can combine `ClipboardCopy` with a link.)

Or this?
image

SpyTec:
> Maybe something like the following to follow GitHub style of showing commits:

Could be possible, I'll draft something up later and show you

SpyTec:
Hah. I drafted a reply and never hit send :D

@lachmanfrantisek WDYT about this?

overall view

existing PR/branch link

new commit link

copy to clipboard

lachmanfrantisek:
@SpyTec Looks really good! Thanks for playing with that!

I would maybe add a bigger space before the commit but that's just a really small nitpick.

SpyTec:
Ack, I'll see about getting that improved!

Will update other pages too

lachmanfrantisek:
Nice, thanks, Eric!

frontend/src/app/Results/ResultsPageTestingFarm.tsx
```diff
@@ -80,98 +164,58 @@ const ResultsPageTestingFarm = () => {
...
-    data.copr_build_ids = data.copr_build_ids.filter((copr) => copr !== null);
-    const coprBuildInfo =
-        data.copr_build_ids.length > 0
-            ? getCoprBuilds(data.copr_build_ids)
-            : "";
+    const onCopyHash = () => {
+        if (data) {
+            navigator.clipboard.writeText(data.commit_sha.substring(0, 7));
+        }
+    };
```
lachmanfrantisek, line 202: What about copying the whole `commit_sha`?

Looks like GitHub uses that as well:
image

lachmanfrantisek:
@SpyTec is this still a draft? I am really looking forward to seeing this merged..;)

SpyTec:
> @SpyTec is this still a draft? I am really looking forward to seeing this merged..;)

@lachmanfrantisek the list of Copr builds needs a slight change but otherwise it's all ready! Feel free to merge

lachmanfrantisek:
> the list of Copr builds needs a slight change but otherwise it's all ready! Feel free to merge

OK, let's merge this as is so we still have some time to verify this on stage and the list of Copr builds can be improved as a follow-up. Thanks a lot for all the work!

lachmanfrantisek (approved):
Thanks Eric! :rocket:

Issue #319, opened 2023-08-15
Crash with TypeError: e.trigger.builds is undefined


danmyway:
Reproducer:
Access https://dashboard.packit.dev/projects/github.com/oamg/convert2rhel and click Load more.

Error message:
TypeError: e.trigger.builds is undefined

lbarcziova:
Thanks for reporting! The cause of this is https://github.com/packit/packit-service/issues/2153 so the progress will be tracked there.

mfocko:
Can we close as the root cause has been resolved?

mfocko:
> What do you mean?

OH! it's the `s.map` from the other issue you just linked

lbarcziova:
The fix is now in production, closing.

PR #320, opened 2023-08-21
Adjust Copr build epoch handling for None

lbarcziova:
Adjust how we include the epoch in the build result page. We get the value in packit-service from Copr API and the behaviour there recently changed: https://github.com/fedora-copr/copr/pull/2876

---

RELEASE NOTES BEGIN
N/A
RELEASE NOTES END


SpyTec (approved):
LGTM

Should fix #319 too

lbarcziova:
recheck

PR #325, opened 2023-09-04
Bump deps

mfocko:
as a follow-up to the vulnerability report


jpopelka (commented):
:confused: 
So, will the GH action for deploying PRs work without it?
(The last run went OK, but the preview is already unavailable)

mfocko:
> 😕 So, will the GH action for deploying PRs work without it? (The last run went OK, but the preview is already unavailable)

yes :) actually mentioned in the commit message, but it didn't get to the PR because of multiple commits :/
I hit some issues when trying to get it to work, so I installed it almost anywhere and I must've missed the commit when dropping the debugging changes, etc. Surge is installed »globally« and »within« the action, so we shouldn't need it in the frontend itself.

mfocko:
https://github.com/packit/dashboard/pull/325/files#diff-da6498268e99511d9ba0df3c13e439d10556a812881c9d03955b2ef7c6c1c655R12-R17

@SpyTec shouldn't those be included only in the dev dependencies? :thinking: we got it tagged as security vulnerability from one of its dependencies (ejs)

jpopelka (approved):

PR #328, opened 2023-09-07
Add usage trend charts for project count

lachmanfrantisek:
This adds two new charts to show the usage of jobs for the number of unique projects.

image

RELEASE NOTES BEGIN
Two new charts are available at https://dashboard.packit.dev/usage to show the trends based on the number of active projects.
RELEASE NOTES END


lbarcziova (approved):
thanks!

lgtm and we can improve later (we discussed in chat that `top_projects_by_job_runs` are upstream URLs at the moment and in the future it would be better to count the dist-git ones)

Issue #306, opened 2023-06-24
Navigation restructure (Dashboard 2.0)


SpyTec:
### Description

Going from packit-service to individual results within the Packit dashboard is great and works as it should. It shows the extend of what is possible from things like Testing Farm

However, if you want to go to the Packit dashboard to view anything else, from a package maintainer there is very little that can be viewed for my project without a lot of hassle or trouble finding them. 

As a Packit maintainer, it is however a perfect view as it shows everything happening regarding Packit and its services

---

For this reason, there is work to be done to make more parts of the application better for both parties, we already have some things planned like 

- #257 
- #196 

But for a completely usable system by package maintainers we need to restructure the navigation and relationship between projects and their builds and runs

### Structure proposal
With a [context selector](https://www.patternfly.org/v4/components/context-selector/design-guidelines) we could put it in the top banner or move it to the top of the sidebar. Within it, we would show projects and allow the user to search for their projects. With a default option being "All projects" which would mean everything functions like how it is now

With the context selected for a specific project, e.g. packit/dashboard
- Jobs would be filtered to only include builds and tests triggered from this repo
- Pipelines would be filtered as well to only include pipelines triggered by this repo
- Projects would show a selection of the current project and allow you to select a different project from there as well

With the context cleared, going for All projects
- Jobs would show all builds and tests as it dows now
- Pipelines would show the latest pipeline from any project
- Projects would show all the projects and allow you to search and select a project for the context

Routes themselves should change with this as well, with previous ones either redirecting or being kept for all-projects

- `/projects` - Current projects list that already exists
- `/projects/packit/dashboard` - Same as above but keeping data about the selected project
- `/projects/packit/dashboard/detail` - the project view itself
- `/projects/packit/dashboard/jobs`
	- `/projects/packit/dashboard/jobs/copr-builds`
	- `/projects/packit/dashboard/jobs/srpm-builds`
	- ...
- `/projects/packit/dashboard/pipelines`

### Workaround

- [ ] There is an existing workaround that can be used until this feature is implemented.

### Participation

- [X] I am willing to submit a pull request for this issue. (Packit team is happy to help!)

SpyTec:
I guess #196 is very similar

lachmanfrantisek:
@SpyTec Thanks for such a detailed description! A few notes:

* I really like being explicit about external links being external...;)
* I quite like that you can still view all the jobs/pipelines but people can filter that.
[...]

PR #336, opened 2023-10-09
refactor: Move sidebar items around

SpyTec:
This will prepare us for more granular interaction with project selection in the future. For now until that is implemented we have this to go with.

image

TODO:

- [ ] ‹fill in›

Fixes

Related to
#306 

Merge before/after

---

RELEASE NOTES BEGIN
Move sidebar routes around and categorize them
RELEASE NOTES END

### DIFF
--- a/frontend/src/app/routes.tsx
+++ b/frontend/src/app/routes.tsx
@@ -39,10 +39,20 @@ const routes: RouteObject[] = [
                 },
             },
             {
+                element: <Projects />,
+                path: "/projects",
+                handle: {
+                    divider: true,
+                    label: "Projects",
+                    category: "Dashboards",
+                },
+            },
+            {
                 Component: Jobs,
                 id: "jobs",
                 path: "/jobs",
                 handle: {
+                    category: "Dashboards",
                     label: "Jobs",
                 },
                 children: [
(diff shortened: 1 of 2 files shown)


frontend/src/app/routes.tsx
```diff
@@ -101,16 +111,10 @@ const routes: RouteObject[] = [
                 element: <Pipelines />,
                 path: "/pipelines",
                 handle: {
+                    category: "Projects",
                     label: "Pipelines",
                 },
             },
```
lachmanfrantisek, line 117: I would group projects, pipelines and jobs, but am not sure if `Projects` should be the name of the group.

But can't think of a better name. Maybe something with results, activity,... I don't know..;)

SpyTec: I was thinking of adding this in the future so we can select the project we want and see the jobs and pipeline just for that project, then if you select none it shows all jobs and all pipelines. Could somehow get a page for just projects somehow though

But we could also leave it for now and just have the external links categorized

248487948-44d41eda-8021-4a8a-8769-4bf84f12e7b2

lachmanfrantisek: OK, now I remember we've discussed this.

I quite like that approach (both having jobs and pipelines in a group now and the future state with the context selector). Maybe, I would just call this group differently.

A few ideas:

* Service
* Packit Service
* Records
* Results
* Dashboard(s)
* Workloads
* Processing

SpyTec: I like both Dashboards and Services

lachmanfrantisek: From these two, I slightly prefer `Dashboards`... Let's pick this and we can easily rename it if needed.

SpyTec: Changed!

SpyTec: image

lachmanfrantisek (approved):
:rocket:

PR #342, opened 2023-10-15
fix: ReleasesList to use PatternFly table

SpyTec:
Currently ReleasesList uses the PatternFly classes for the table but not
the table itself. This changes it to use the PF component

TODO:

- [ ] ‹fill in›

Fixes

Related to

Merge before/after

---

RELEASE NOTES BEGIN
Change releases table in Project details to use PatternFly table
RELEASE NOTES END


lachmanfrantisek (approved):
:+1: 

Now, we should put some relevant info into the table..;)

PR #344, opened 2023-10-15
fix: External link in TriggerInfo to show link

SpyTec:
Before this change it opened the link in a new page/tab but did not
show the URL it is going to, so you could not copy the link or see that
it is a link

TODO:

- [ ] ‹fill in›

Fixes

Related to

Merge before/after

---

RELEASE NOTES BEGIN
Show the external links correctly in project view
RELEASE NOTES END


lachmanfrantisek (approved):
Thanks!

PR #345, opened 2023-10-15
fix: correct labels for Projects in small viewport

SpyTec:
Within the Projects view the rows when in a small viewport does not
display the correct labels. This is now corrected

Before:
before

After:
after

TODO:

- [ ] ‹fill in›

Fixes

Related to

Merge before/after

---

RELEASE NOTES BEGIN
Address incorrect label and label format in Projects view
RELEASE NOTES END


lachmanfrantisek (approved):
Well spotted!

PR #354, opened 2023-11-08
Show downstream Koji builds

### LINKED ISSUE #214: Show downstream Koji builds
* [ ] Make sure that detail pages correctly show results for downstream Koji builds.
    * [ ] differentiate downstream and upstream builds
* [ ] Make sure the downstream Koji builds are correctly shown in the jobs view: https://dashboard.packit.dev/jobs
    * [ ] differentiate downstream and upstream builds
* [ ] Make sure the downstream Koji builds are correctly shown in the pipelines view: https://dashboard.packit.dev/pipelines
    * [ ] differentiate downstream and upstream builds

Requires: packit/packit-service#1889

---
*Part of packit/dashboard#187 epic.*

lbarcziova:
Create a separate table for downstream Koji builds and adjust the result view to the API changes.

Fixes #214

Snímka obrazovky z 2023-11-08 11-04-46

Snímka obrazovky z 2023-11-09 14-55-15

---
RELEASE NOTES BEGIN
Downstream Koji builds are now shown in a separate table in `/jobs/downstream-koji-builds`.
RELEASE NOTES END

### DIFF
--- a/frontend/src/app/Jobs/Jobs.tsx
+++ b/frontend/src/app/Jobs/Jobs.tsx
@@ -95,6 +95,16 @@ const Jobs = () => {
                                     Pull From Upstreams
                                 </NavLink>
                             </NavItem>
+                            <NavItem
+                                isActive={
+                                    currentMatch?.id ===
+                                    "downstream-koji-builds"
+                                }
+                            >
+                                <NavLink to={"downstream-koji-builds"}>
+                                    Downstream Koji Builds
+                                </NavLink>
+                            </NavItem>
                         </NavList>
                     </Nav>
                 </PageNavigation>
--- a/frontend/src/app/Jobs/KojiBuildsTable.tsx
+++ b/frontend/src/app/Jobs/KojiBuildsTable.tsx
@@ -18,12 +18,12 @@ import { useInfiniteQuery } from "@tanstack/react-query";
 
 export interface KojiBuild {
     packit_id: number;
-    build_id: string;
+    task_id: string;
     status: string; // TODO(SpyTec): Probably an enum right? Change to be one if so
     build_submitted_time: number;
     chroot: string;
     web_url: string;
-    build_logs_url: string;
+    build_logs_urls: string;
     // TODO(SpyTec): change interface depending on status of pr_id or branch_item.
     // They seem to be mutually exclusive so can be sure one is null and other is string
     pr_id: number | null;
@@ -33,8 +33,10 @@ export interface KojiBuild {
     repo_namespace: string;
     repo_name: string;
 }
-
-const KojiBuildsTable = () => {
+interface KojiBuildTableProps {
+    scratch: "true" | "false";
+}
+const KojiBuildsTable: React.FC<KojiBuildTableProps> = ({ scratch }) => {
     // Headings
     const columns = [
         {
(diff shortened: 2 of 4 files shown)


SpyTec:
I wonder, should we change it to be Koji builds and have a filter dropdown inside that component for upstream vs downstream?

lbarcziova:
@SpyTec good point, I was thinking about it as well but in the end decided to go with a separate table so that each table corresponds to a CI job that can be configured by a user. It is a similar situation that with `Propose Downstreams`/`Pull From Upstreams`.  @packit/the-packit-team any opinions?

SpyTec:
With propose downstreams and pull from upstream, if we consolidated those two what would we name it?

lbarcziova:
@SpyTec probably `Release syncs` or similar.

majamassarini:
> @SpyTec good point, I was thinking about it as well but in the end decided to go with a separate table so that each table corresponds to a CI job that can be configured by a user. It is a similar situation that with `Propose Downstreams`/`Pull From Upstreams`. @packit/the-packit-team any opinions?

I think we don't have so many jobs at the moment and it is nice to get the result with one less click so I would let them as they are at least for now.
If possible, I would show the Koji build related tabs one near the other.

SpyTec:
> I think we don't have so many jobs at the moment and it is nice to get the result with one less click so I would let them as they are at least for now.

@majamassarini totally fair! I would like to highlight however that it is a lot of words for reach tab and it becomes a lot to read. There is no clear or easy distinction between the tabs apart from the text so it is difficult to find fast. Hence why I think that consolidating it to one page would be better as it reduced the amount of tabs while also making it easier to find the tab you're looking for (as it would be "Copr Builds" only

majamassarini:
> > I think we don't have so many jobs at the moment and it is nice to get the result with one less click so I would let them as they are at least for now.
> 
> @majamassarini totally fair! I would like to highlight however that it is a lot of words for reach tab and it becomes a lot to read. There is no clear or easy distinction between the tabs apart from the text so it is difficult to find fast. Hence why I think that consolidating it to one page would be better as it reduced the amount of tabs while also making it easier to find the tab you're looking for (as it would be "Copr Builds" only

I see your point and I have the bias to use it every day, so I am not a good user example 😅 So +1 for consolidating them.

lbarcziova:
We have discussed this in our architecture meeting and agreed on either having all the views separately or having both Koji builds and Release syncs consolidated together. I would probably merge this PR with this initial implementation and in the future, we could improve this by consolidating the views and using filters while preserving the paths for the views.

lbarcziova:
I would say they should, I was also thinking about `upstream/downstream`, but preferred this one since it is more correct.

lachmanfrantisek:
> Will people know what scratch and non-scratch is?

I was thinking about the same -- `scratch` is fine, but not sure about `non-scratch`.  Maybe `production`, or don't mention anything for downstream ones?

lbarcziova:
ok, I would then go with `scratch` and `production`

TomasTomecek (approved):
LGTM, I like the proposed design as it's more straightforward and intuitive with less clicks to see what I want.

At the same time, agreed with @SpyTec that the tab line is getting pretty long and we should think about a long term solution: the number of jobs will only grow :)

PR #371, opened 2024-01-24
fix(ResultsPageSRPM): fix missing link when SRPM build fails

jamacku:
When the RPM build fails during the creation of SRPM, the `Logs` link is missing.

e.g. https://dashboard.packit.dev/results/copr-builds/1279980

Note: This is probably not the correct solution.

---

TODO:

- [ ] ‹fill in›

Fixes

Related to

Merge before/after

---

RELEASE NOTES BEGIN
￼
RELEASE NOTES END

### DIFF
--- a/frontend/src/app/Results/ResultsPageSRPM.tsx
+++ b/frontend/src/app/Results/ResultsPageSRPM.tsx
@@ -171,7 +171,7 @@ const ResultsPageSRPM = () => {
                       {" "}
                       (
                       <a
-                        href={data.logs_url}
+                        href={data.logs_url ? data.logs_url : `${URL}/builder-live.log`}
                         rel="noreferrer"
                         target={"_blank"}
                       >
--- a/frontend/src/app/Results/ResultsPageSRPM.tsx
+++ b/frontend/src/app/Results/ResultsPageSRPM.tsx
@@ -171,7 +171,11 @@ const ResultsPageSRPM = () => {
                       {" "}
                       (
                       <a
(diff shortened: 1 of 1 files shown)


SpyTec:
Thanks! Is there a logfile anywhere we can use instead? Otherwise we could hide the `(Logs)` part entirely

Currently the URL you used would point to https://dashboard.packit.dev/results/copr-builds/1279980/builder-live.log which isn't valid

jamacku:
This is the right URL: https://download.copr.fedorainfracloud.org/results/packit/jamacku-linux-pam-1/srpm-builds/06946722/builder-live.log.gz

I have misunderstand what ${URL} is :)

SpyTec:
Gotcha! Hmm, do we want to add a new field to the API endpoint to return the logfile for SRPM as well? @lbarcziova WDYT should be the ideal here?

lbarcziova:
Thanks @jamacku for the contribution and @SpyTec for the review! 

If I check correctly, this should be the same issue as packit/packit-service#2195, right? I think we wanted to actually not show these Copr jobs in dashboard at all (only the SRPM ones), would that work for you or do you prefer having it in dashboard with fixed link?

jamacku:
By the way, the logs are still accessible from the copr UI with just two clicks. So, alternatively, you can hide the link if the URL is missing.

Venefilyn:
I just saw it by accident of what you meant and yeah it makes sense to show it. I've updated SRPM Build pages to always include the build logs in #440. That is unless I'm mistaken in how Packit service determines that link, but should be fixed once that's merged!

PR #373, opened 2024-02-06
Add goal Q1 2024 to usage/total charts

majamassarini:
Fixes packit/packit-service#2304

Merge after packit/packit-service#2330

### DIFF
--- a/frontend/src/app/Usage/UsageList.tsx
+++ b/frontend/src/app/Usage/UsageList.tsx
@@ -335,7 +335,6 @@ const UsageList: React.FC<UsageListProps> = (props) => {
           <Flex>{job_charts}</Flex>
         </CardBody>
       </Card>
-      {getGoalProgress()}
       {getGoalProgress2024Q1()}
     </>
   );
--- a/frontend/src/app/Usage/UsageList.tsx
+++ b/frontend/src/app/Usage/UsageList.tsx
@@ -254,6 +254,45 @@ const UsageList: React.FC<UsageListProps> = (props) => {
     );
   }
 
+  function getGoalProgress2024Q1() {
+    if (props.what != "total") {
+      return <></>;
+    }
+    return (
+      <Card>
+        <CardTitle>Packit goal for Q1/2024</CardTitle>
+        <CardBody>
+          <Flex>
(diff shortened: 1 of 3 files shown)


majamassarini:
recheck

lachmanfrantisek (commented):
Thanks a lot, Maja! When you are in this, can you please remove the chart from the last quarter?

lbarcziova (approved):
:+1: 

do you think we should also show the almost_onboarded part? :thinking:

frontend/src/app/Usage/UsageList.tsx
```diff
@@ -254,6 +254,45 @@ const UsageList: React.FC<UsageListProps> = (props) => {
...
+                  ariaDesc="Chart describing downstream onboarding."
+                  ariaTitle="Chart describing downstream onboarding."
+                  constrainToVisibleArea
+                  labels={({ datum }) => `${datum.name}: ${datum.y}`}
+                  maxDomain={{ y: 500 }}
+                  name="Chart describing downstream onboarding."
+                  primarySegmentedMeasureData={[
+                    {
+                      name: "Onboarded projects",
+                      y: data.onboarded_projects_q1_2024,
```
lbarcziova, line 281: actually, where is this field coming from please?

majamassarini:
> 👍
> 
> do you think we should also show the almost_onboarded part? 🤔

Do you mean the number of the almost_onboarded projects or their url? 
I think the useful part is their url, so we can maybe do something to push them forward, but I don't think this view is the right place where to have this list. And it is so easy to get it via the usage page (for us) that I would not spend time on it. But probably this is also because I don't know React so well to do it quickly...

frontend/src/app/Usage/UsageList.tsx
majamassarini: https://github.com/packit/packit-service/blob/48e820856a32aa1f997324bafcec3eec865f67a3/packit_service/service/api/usage.py#L738

lbarcziova: sorry it looks like I am blind :D thanks for the link!

PR #385, opened 2024-03-22
Change Prettier formatter tabbing from 4 to 2

SpyTec:
This changes the default tabbing to be 2 spaces instead of the current
4 spaces.

tsconfig.json is also changed to add missing commas

TODO:

- [ ] ‹fill in›

Fixes

Related to

Merge before/after

---

RELEASE NOTES BEGIN
￼
RELEASE NOTES END

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -71,7 +71,7 @@
           "*.(js|jsx|ts|tsx)"
         ],
         "options": {
-          "tabWidth": 4
+          "tabWidth": 2
         }
       }
     ]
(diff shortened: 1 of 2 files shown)


frontend/package.json
```diff
@@ -71,7 +71,7 @@
           "*.(js|jsx|ts|tsx)"
         ],
         "options": {
-          "tabWidth": 4
+          "tabWidth": 2
```
lachmanfrantisek, line 74: Is there a reason for this? Isn't there any "default" or usually-set value around the community?


I am quite used to the 4 from the Python world... But am not familiar with how it's done elsewhere...

SpyTec: Default is usually 2 in the JavaScript world. Most of our files are also tabbed with 2 spaces already, so this just preserves that

lbarcziova (approved):

PR #377, opened 2024-02-14
feat: Add pipeline detail page

SpyTec:
This adds a new pipeline details page that shows all the
statuses for the different jobs run in the pipeline

There can be a lot of future improvements to this, such as
inline fetching of the different resources and generally including
more fields to the merged pipeline results, however for a MVP this
is as good as can be while being as simple as possible
TODO:

- [ ] ‹fill in›

Fixes

Related to

Merge after
* #386 

---

RELEASE NOTES BEGIN
Add pipeline details page
RELEASE NOTES END

PR #386, opened 2024-03-22
refactor(TriggerLink): Split link and text

SpyTec:
This splits the linking from the text in the TriggerLink component
so that it now can be used in a more versatile way, such as setting
internal Packit dashboard links with the display of TriggerLink.

As UX is going to be more refined in the future, this also helps pay the
way of changing external links to be more uniform without having to
change the TriggerLink once again.

For what is being worked on at the moment with getting Pipeline details
this change would be most welcome when it comes to linking internally
but having the text from TriggerLink.

Related
#377 

TODO:

- [ ] ‹fill in›

Fixes

Related to

Merge before/after

---

RELEASE NOTES BEGIN
Split TriggerLink into two components to make UX improvements easier
RELEASE NOTES END

### DIFF
--- a/frontend/src/app/Jobs/BodhiUpdatesTable.tsx
+++ b/frontend/src/app/Jobs/BodhiUpdatesTable.tsx
@@ -11,7 +11,7 @@ import {
 } from "@patternfly/react-table/deprecated";
 
 import { Button } from "@patternfly/react-core";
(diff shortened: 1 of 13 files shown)


SpyTec:
Moved this change out from #377 and rebased that one to this

lbarcziova (approved):
lgtm, thanks!

PR #391, opened 2024-04-05
fix: Redirect jobs page better

SpyTec:
Whenever you went to the Jobs page from the sidebar, you could click jobs again and it didn't redirect you to the copr-builds anymore.

This meant that the page would display no tabs selected and there would be no content below the tabs, such as the Copr Builds table. To fix this we had to fix the `useEffect` to include `matches` which has the URL information. That way whenever the URL changes this will also change

To make it even nicer on the eyes, we will hard-import Copr Builds table so that it will be visible if the url is just `/jobs` again and it didn't redirect correctly. This also makes it so that the rendering of the table won't flicker if you click Jobs in the sidebar over and over. Note that the statuses for the table will still flicker, as it loads the table again and the implementation as of this commit does not render the statuses properly and unnecessarily renders it (AFAIK).

TODO:

- [ ] ‹fill in›

Fixes

Related to

Merge before/after

RELEASE NOTES BEGIN

RELEASE NOTES END

### DIFF
--- a/frontend/src/app/Jobs/Jobs.tsx
+++ b/frontend/src/app/Jobs/Jobs.tsx
@@ -12,7 +12,7 @@ import {
   Text,
   TextContent,
 } from "@patternfly/react-core";
-import { useEffect } from "react";
+import { useEffect, useMemo } from "react";
 import {
   NavLink,
   Outlet,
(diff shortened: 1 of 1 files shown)


lachmanfrantisek (approved):
Thanks!

PR #394, opened 2024-04-07
feat(UX): Step timeline for SRPM build status

SpyTec:
Adds the step pipeline we use in Copr Builds to SRPM Builds page as well

TODO:

- [ ] ‹fill in›

Fixes

Related to

Merge before/after

RELEASE NOTES BEGIN
Add step timeline for SRPM build status
RELEASE NOTES END

### DIFF
--- a/frontend/src/app/Results/ResultsPageSRPM.tsx
+++ b/frontend/src/app/Results/ResultsPageSRPM.tsx
@@ -29,6 +29,7 @@ import { Timestamp } from "../utils/Timestamp";
 import { useParams } from "react-router-dom";
 import { useTitle } from "../utils/useTitle";
 import { useQuery } from "@tanstack/react-query";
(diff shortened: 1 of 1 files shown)


SpyTec:
> Would it make sense to make the timeline step clickable and get rid of the `pending` label?

@lachmanfrantisek for a future suggestion I'd say possibly but that depends on how well it aligns with other jobs. But I wouldn't do it for MVP

It's better to have it like this for now than have SRPM builds be different from the other jobs

lbarcziova (approved):

lachmanfrantisek:
> It's better to have it like this for now than have SRPM builds be different from the other jobs

I meant it also in general for all the jobs that we would show the state two times:
image

But if you think it's easier to fix later than now, before merging PRs for all the jobs, we can merge it.

PR #410, opened 2024-05-21
Feat/sentry-integration

Venefilyn:
This adds Sentry integration to the dashboard which will report any
errors or issues. In the future, we can add a modal as well where the
user can provide their own feedback.

TODO:

- [ ] ‹fill in›

Fixes

Related to

Merge before/after

RELEASE NOTES BEGIN

RELEASE NOTES END


lbarcziova (commented):
LGTM, thanks for adding this!

I would maybe still be for not having the `dsn` directly here, @mf WDYT?

Venefilyn:
@lbarcziova I changed it. If you want we can rotate the dsn as well but I doubt it'll be an issue

lbarcziova (approved):
thank you so much! We do it this way also in the workers and service, therefore I was inclined to have it here like that as well.

Venefilyn:
@lbarcziova Np! Feel free to merge when the environment variables are in the CI/deployment process for stg prod

PR #413, opened 2024-05-29
Expose SENTRY_AUTH_TOKEN for frontend build

lbarcziova:
Followup of #410

I would like to avoid writing the secret into a file, but the `buildah` version in the action doesn't support env vars. Therefore open to any suggestions and opening it as draft :pray:

### DIFF
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -42,6 +42,11 @@ jobs:
           echo "commit_sha=${GITHUB_SHA::7}" >> $GITHUB_OUTPUT
         id: branch_tag
 
+      - name: Set up secret file needed for the build
+        run: echo "$SENTRY_AUTH_TOKEN" > /tmp/sentry-secret
+        env:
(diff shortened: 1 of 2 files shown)


Venefilyn (approved):
I looked things up and didn't see anything that we could utilize to make this foolproof. But no one would be able to get the secret that easily luckily. It's only something that can be accessed from within the repo

As for the secret itself, when we add it we should make sure it's unique to this specific repo and not a generic Packit one that is used everywhere. Means less work if it ends up leaked in a log somewhere as we just have to disable it for this repo

.github/workflows/build-and-push.yml
```diff
@@ -42,6 +42,11 @@ jobs:
           echo "commit_sha=${GITHUB_SHA::7}" >> $GITHUB_OUTPUT
         id: branch_tag
 
+      - name: Set up secret file needed for the build
+        run: echo "$SENTRY_AUTH_TOKEN" > /tmp/sentry-secret
```
mfocko, line 46: So that's the reason why GitLab allows to have secrets mounted as files :D

mfocko (approved):

PR #417, opened 2024-06-05
feat: Pipeline detail view

Venefilyn:
Commit message to be changed, still in draft.

Results so far
image

TODO:

- [ ] ‹fill in›

Fixes

Related to

Merge before/after

RELEASE NOTES BEGIN
Introduce detail page for individual pipelines
RELEASE NOTES END

### DIFF
--- a/frontend/src/app/Pipelines/PipelineDetail.tsx
+++ b/frontend/src/app/Pipelines/PipelineDetail.tsx
@@ -13,8 +13,6 @@ import {
   DescriptionListGroup,
   DescriptionListTerm,
   LabelGroup,
-  List,
-  ListItem,
   PageSection,
   PageSectionVariants,
   Skeleton,
@@ -28,6 +26,7 @@ import kojiLogo from "../../static/koji.ico";
 import { useQuery } from "@tanstack/react-query";
 import { useParams } from "react-router";
 import { TriggerLink, TriggerSuffix } from "../Trigger/TriggerLink";
+import { Timestamp } from "../utils/Timestamp";
 
 interface StatusItem {
   packit_id: number;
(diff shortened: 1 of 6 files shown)


frontend/src/app/Pipelines/PipelinesTable.tsx
```diff
@@ -151,9 +162,9 @@ const PipelinesTable = () => {
           {
             title: (
               <strong>
-                <TriggerLink trigger={run.trigger}>
+                <Link to={`/pipelines/${run.merged_run_id}`}>
                   <TriggerSuffix trigger={run.trigger} />
-                </TriggerLink>
+                </Link>
```
lbarcziova, line 167: maybe this is because I am already used to how it works currently (that it links the event e.g. on GitHub), but I am wondering whether it wouldn't be better to have it as a separate new link, e.g. `Pipeline <pipeline-id>` in the beginning, or `Pipeline details` in the end of the row, WDYT?

lbarcziova (commented):
looks great so far!

frontend/src/app/Pipelines/PipelinesTable.tsx
Venefilyn: UX wise it would be better to have it at the end. But yes I'll add a link to the end or something. External links should be treated secondary to in app links

Venefilyn: I changed it to look like this
image

lbarcziova: great, this looks good :+1:

frontend/src/app/Pipelines/PipelineDetail.tsx
```diff
@@ -0,0 +1,318 @@
...
+      {icon}&nbsp;<span>{text}</span>
+    </>
+  );
+}
+
+const PipelineDetail = () => {
+  let { id } = useParams();
+
+  const fetchData = () =>
+    fetch(`${import.meta.env.VITE_API_URL}/runs/merged/${id}`).then(
```
lbarcziova, line 129: I think there is some issue with this API endpoint, because for some of the pipelines, I got different pipeline in the pipelineview than I clicked on in the table..I will have a look into it later

lbarcziova (commented):
visually looks great!

frontend/src/app/Pipelines/PipelineDetail.tsx
Venefilyn: I wonder if there's an issue with cache. Does it happen on the api endpoint?

lbarcziova: I was able to find such entries also via API directly :/

lbarcziova: I hope packit/packit-service#2452 fixed this

Venefilyn:
@lbarcziova something like this?

image
image

PR #418, opened 2024-06-10
chore: remove Prettier from pre-commit config

Venefilyn:
Pre-commit prettier is not maintained anymore and if we want to run it
we should consider an alternative way

https://github.com/pre-commit/mirrors-prettier

RELEASE NOTES BEGIN
Remove Prettier from pre-commit
RELEASE NOTES END

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,11 +8,6 @@ repos:
     hooks:
       - id: black
         language_version: python3
-  - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: v4.0.0-alpha.8
-    hooks:
-      - id: prettier
-        exclude: frontend/tsconfig.json
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.6.0
     hooks:


Venefilyn:
> thanks for noticing and opening the PR! We are using it across multiple other `packit` repositories, so we should consider using some other fork or some alternative

@lbarcziova could be solved by dprint #419. They don't have pre-commit but they have a GH action

Venefilyn:
Okay [Biome](https://biomejs.dev/) is more popular and has a bit better maintainability for now. I'd vote in favor of that. In #447 currently I run into an issue again with prettier wanting to format a file and we can't fix it easily using pre-commit. I'd rather go to using husky or just a GitHub action for handling formatting using something like Biome

Where else is prettier used atm in other projects? It should only be a JS formatter @lbarcziova

lbarcziova:
@Venefilyn we use prettier in other projects to format Markdown, yaml and JSON (e.g. test response files). 

Biome looks cool as a JS formatted, if I checked correctly it doesn't support yaml and Markdown, right?

Venefilyn:
> @Venefilyn we use prettier in other projects to format Markdown, yaml and JSON (e.g. test response files).
> 
> Biome looks cool as a JS formatter, if I checked correctly it doesn't support yaml and Markdown, right?

@lbarcziova you're right, currently it doesn't support it https://biomejs.dev/internals/language-support/

PR #426, opened 2024-06-25
perf: Minimize images as they are used in the app

Venefilyn:
We have the images already in the source code but we do not minimize
depending on where we use them. This takes care of that

We also now split the JS files a bit to improve fetching, but it is not
a performance improvement as they all will still load at the same time

TODO:

- [ ] ‹fill in›

Fixes

Related to

Merge before/after

RELEASE NOTES BEGIN
Automatically reduce image size on build
RELEASE NOTES END


lbarcziova (approved):
:+1:

PR #429, opened 2024-06-26
chore: change from yarn to pnpm

Venefilyn:
This changes the package manager for the dashboard dependencies over
for the reason that pnpm is a bit more uniformly supported and has
more features. pnpm also takes up less storage and is faster at
installing.

We can also scrap this idea and go with installing and using https://bun.sh/
instead of pnpm

TODO:

- [ ] ‹fill in›

Fixes

Related to

Merge before/after

RELEASE NOTES BEGIN
Change package manager from yarn to pnpm
RELEASE NOTES END


frontend/package.json
```diff
@@ -154,5 +154,22 @@
...
+      "os": [
+        "win32",
+        "darwin",
+        "current"
+      ],
+      "cpu": [
+        "x64",
+        "arm64"
+      ]
+    }
```
Venefilyn, line 169: Something like this is needed for sharp, which optimizes the images for us

https://github.com/packit/dashboard/actions/runs/9682145558/job/26714460555

Venefilyn:
@lbarcziova WDYT of using something like this? I know I've talked a bit with Miro who brought up Bun as well

lbarcziova:
@Venefilyn I will have a look into those as I am not that familiar with them and let you know! Thanks for the ideas!

Venefilyn:
> @Venefilyn I will have a look into those as I am not that familiar with them and let you know! Thanks for the ideas!

@lbarcziova did you get a chance to look at the current options out there? IIRC just yarn, npm, pnpm, bun

lbarcziova:
@Venefilyn yes, I looked into a few comparisons, and if the decision would be pnpm vs bun, for now, I would go with pnpm. I see that pnpm can have some advantages compared to yarn, I am just not sure how much benefit it would bring for our use case at the moment. But since you are more experienced in these technologies, if you think this would be worth the transition, we can go for it!

Venefilyn:
Primarily for now it would help with sharp as it doesn't work very well with yarn as far as I can see, and there are some issues with the different yarn lock file versions I don't like. This might need containers to be updated in base Packit repos but unsure

Anyway, I think it's ready for merge

lbarcziova (approved):
thanks, let's try this!

PR #431, opened 2024-06-27
Move react-query-devtools from devDependencies

lbarcziova:
Otherwise the production build fails.


lbarcziova:
:facepalm: thanks, closing :smile:

PR #434, opened 2024-07-02
Fix Copr build installation instructions

nforro:
`dnf` is in EL since EL8, `yum` there is just a compatibility wrapper. Also the subsequent steps don't consider `yum` at all.
Just get rid of it.

### DIFF
--- a/frontend/src/app/Results/ResultsPageCopr.tsx
+++ b/frontend/src/app/Results/ResultsPageCopr.tsx
@@ -149,11 +149,7 @@ const ResultsPageCopr = () => {
             <br />
             <List>
               <ListItem>
-                <code>sudo yum install -y dnf-plugins-core</code> on RHEL 8 or
-                CentOS Stream
-              </ListItem>
-              <ListItem>
-                <code>sudo dnf install -y dnf-plugins-core</code> on Fedora
+                <code>sudo dnf install -y dnf-plugins-core</code>
               </ListItem>
               <ListItem>
                 <code>


frontend/src/app/Results/ResultsPageCopr.tsx
```diff
@@ -149,11 +149,7 @@ const ResultsPageCopr = () => {
             <br />
             <List>
               <ListItem>
-                <code>sudo yum install -y dnf-plugins-core</code> on RHEL 8 or
-                CentOS Stream
-              </ListItem>
-              <ListItem>
-                <code>sudo dnf install -y dnf-plugins-core</code> on Fedora
+                <code>sudo dnf install -y dnf-plugins-core</code>
```
nforro, line 152: We could also be more specific, WDYT?

```suggestion
                <code>sudo dnf install -y 'dnf-command(copr)'</code>
```

majamassarini: I am not sure, reading [here](https://docs.pagure.org/copr.copr/how_to_enable_repo.html#how-to-enable-repo) I will stick with `sudo dnf install -y dnf-plugins-core`

majamassarini (approved):
🙏🏻

frontend/src/app/Results/ResultsPageCopr.tsx
nforro: The benefit of using the virtual provide is that if the package is renamed or split into more packages for example it will still work.

PR #435, opened 2024-07-02
feat: Migrate routing from React Router to TanStack Router

Venefilyn:
This is part of the huge redesign of the frontend website where we will separate routing from components in an easier way to allow us to find routes a lot easier within the codebase. As a bonus with this approach is that we can also do lazy loading and speed up loading times

As an example, see files like `usage.lazy` that is now loaded separately when navigating to it because of lazy loading. That means the application will not load resources if they're not needed. As we do not use `@patternfly/react-charts` everywhere we can also reduce dependency loading

In this refactor a few things have been done
- We have moved routing from a `routes.tsx` format to a file-based format where each file has to declare it is a route and what the name is. -- This also means we will no longer generate the sidebar, but that will have its benefits as well.
- `/app` is removed in favor of routes and components
- There is also a components/ directory for handling displaying more complex routes so there is a separation of concern.
- Multiple tables have been upgraded from deprecated PatternFly Table to their successor Table component. With this I also added skeleton loaders (so it is less jarring on the eyes)
- API queries are now defined in `/queries` directory which allows us to use it anywhere, separation of concern
- API definitions are now centralized in `apiDefinitions.ts`
- Routing changes
  - `/results` has been renamed to be part of `/jobs` -- redirect exists
  - `/jobs`:
	  - `/bodhi-updates` has been renamed to `/bodhi` -- redirect exists
	  - `/copr-builds` has been renamed to `/copr` -- redirect exists
	  - `/downstream-koji-builds` has been renamed to `/koji-downstream` -- redirect exists
	  - `/koji-builds` has been renamed to `/koji` -- redirect exists
	  - `/propose-downstreams` has been renamed to `/sync-release/downstream` -- redirect exists
	  - `/pull-from-upstreams` has been renamed to `/sync-release/upstream` -- redirect exists
	  - `/srpm-builds` has been renamed to `/srpm` -- redirect exists
	  - `/testing-farm-runs` has been renamed to `/testing farm` -- redirect exists
  - `/pipelines` has been renamed to `/pipeline` -- redirect exists

With this directory structure we can also enforce ourselves to remove helpers from components and routes as they will not fit the directory name. Instead we will need to think of something new

This has full feature parity with existing codebase where nothing should break

https://tanstack.com/router/latest

---

During draft form not all routes are converted. Instead they will be converted over time

RELEASE NOTES BEGIN
Enhance routing logic and increase performance
RELEASE NOTES END


frontend/src/routes/jobs/downstream-koji-builds.tsx
```diff
@@ -0,0 +1,15 @@
...
+// SPDX-License-Identifier: MIT
+
+import { createFileRoute, redirect } from "@tanstack/react-router";
+
+export const Route = createFileRoute("/jobs/downstream-koji-builds")({
+  loader: ({ location }) => {
+    throw redirect({
+      to: location.pathname.replace(
+        "/downstream-koji-builds",
+        "/koji/downstream",
```
lbarcziova, line 11: this currently doesn't exist, so it crashes (it should be `/koji-downstream`)

frontend/src/routes/pipeline_.$id.tsx
lbarcziova, line 1: I am thinking about the structure of `routes` directory, would it make sense to rename `jobs_` to something like `job_items` or `items` and then maybe include here also `pipeline_$id.tsx`? Or is the naming somehow restricted?

frontend/src/components/pipeline/PipelinesTable.tsx
```diff
@@ -0,0 +1,259 @@
...
+          )}{" "}
+          <Link to={`/pipeline/${run.merged_run_id}`}>
+            {"git_repo" in run.trigger ? (
+              <>
+                <TriggerSuffix trigger={run.trigger} />
+              </>
+            ) : (
+              <>{run.merged_run_id}</>
+            )}
+          </Link>
```
lbarcziova, line 200: I think there is some issue with this code because this is how pipelines are loaded for me when trying locally:

Snímka obrazovky z 2024-08-14 15-49-38

lbarcziova (commented):
thanks a lot, this must have been a lot of work, we very much appreciate it! :pray: 

I will try to play with it a little bit more locally

as for the routing changes, I am thinking how to make some things more consistent and clear, mainly:
- if we have 2 levels for sync-release, whether we should not do the same for koji-builds
- for sync-release, I am actually thinking this is currently not very clear, I would either use the whole name after `/`, e.g. `sync-release/pull-from-upstream`, or `sync-release/from-upstream` and `sync-release/downstream`

frontend/src/components/pipeline/PipelinesTable.tsx
Venefilyn: Found the issue, you merged this commit yesterday that removed `git_repo`. I had to have the check here because some jobs do not have a git_repo linked, IIRC propose downstream stuff. What should we use now? 

https://github.com/packit/packit-service/commit/1996b57d5f67abf46b7655ce6ac8fed02140eb8d

Note this is also the kind of changes I want to avoid without changing API responses so nothing breaks, as this is a major change in API terms so it warrants a major version upgrade :D

frontend/src/routes/jobs/downstream-koji-builds.tsx
Venefilyn: Ahh good point! I'll leave it until we figure out what to do with Koji upstream and downstream links. `/koji/downstream` is what I tried before but turned out to not work when we had `/koji/$id`

frontend/src/components/pipeline/PipelinesTable.tsx
lbarcziova: Oh, I haven't realised that's the cause, I am currently working on the dashboard follow-up for that PR (as written in [here](https://github.com/packit/packit-service/pull/2498#issue-2457904118). The change intended to actually unify the naming, because somewhere we used `git_repo` and somewhere `project_url` for the same attribute, so now it is `project_url` everywhere.

Venefilyn: Sounds good, I can change this to check for `project_url` instead, though I do advocate for stronger API versioning standards with major/breaking changes like this :D

Venefilyn:
@lbarcziova I can change it to be `jobs/pull-from-upstream/$id` and so which would be good enough for now. But I'll wait to hear what you say from the meeting tomorrow. It's not too much work :D

lbarcziova:
@Venefilyn we have briefly discussed and agreed that using just `(from-)upstream`/`downstream` might cause more confusion, so would rather stick to including the whole job name for the release sync jobs.

Venefilyn:
> @Venefilyn we have briefly discussed and agreed that using just `(from-)upstream`/`downstream` might cause more confusion, so would rather stick to including the whole job name for the release sync jobs.

@lbarcziova so in short, `/sync-release/from-upstream` and `/sync-release/propose-downstream`? Or just using what is currently live, as in `/pull-from-upstream` and `/propose-downstream`?

lbarcziova:
Using what is currently live :+1:

frontend/src/routes/pipeline_.$id.tsx
lbarcziova: ah I see, ok then, thanks for the explanation

Venefilyn:
In that case the PR is ready to be merged if testing and structure is approved

lbarcziova (approved):
LGTM, could you please just change [this](https://github.com/packit/dashboard/pull/435#discussion_r1717114080) so it doesn't crash?

Venefilyn:
> LGTM, could you please just change [this](https://github.com/packit/dashboard/pull/435#discussion_r1717114080) so it doesn't crash?

Already fixed! Just forgot to mark as resolved

Venefilyn:
@lbarcziova fixed in 4c59ac624f2ff8d63d9dd473a7a7d7558d8b037d, also added the router devtools in development, checked and it doesn't show in production

merge whenever!

lbarcziova:
@Venefilyn thanks, merging!

PR #438, opened 2024-07-29
Show multiple NVRs on Bodhi update result page

nforro:
This will show a space separated list, but perhaps having a NVR per line would be better? If so, I welcome any tips how to do that with PatternFly :slightly_smiling_face: 

Related to https://github.com/packit/packit-service/issues/2379.

Merge after https://github.com/packit/packit-service/pull/2475.


Venefilyn (changes requested):
Update the `interface BodhiUpdate` as well to contain `koji_nvrs` (and `koji_nvr` if that is still gonna exist)

Venefilyn (approved):

PR #440, opened 2024-08-19
fix: Always show build log link for SRPM

Venefilyn:
This makes the build log always visible as it will always return from
packit. Otherwise users would not get any logs when an SRPM build has
failed, meaning they would have to chase it down through the link

Takes up on results from #371

TODO:

- [ ] ‹fill in›

Fixes

Related to

Merge before/after

RELEASE NOTES BEGIN
Always show SRPM Build logs
RELEASE NOTES END


lbarcziova (commented):
could we check for `null` for `logs_url` as well? In case there is failure on build submission

Venefilyn:
> could we check for `null` for `logs_url` as well? In case there is failure on build submission

Done!

lbarcziova (approved):
thanks!

PR #442, opened 2024-08-21
fix: add parenthesis back to SRPM Build log links

Venefilyn:
In a previous PR I removed the parenthesis as part of the refactor
without noticing. This commit adds it back

image

RELEASE NOTES BEGIN
Fix Build logs and Results for SRPM builds not being in parenthesis
RELEASE NOTES END


lbarcziova (approved):
thanks!

PR #446, opened 2024-09-02
Skip pipelines without triggers from displaying

lbarcziova:


Venefilyn (commented):
Feels strange that triggers can be undefined. Could you update the `apiDefinitions.ts` as well? It might catch a few more errors we should catch, like in the pipeline detail

frontend/src/components/pipeline/PipelinesTable.tsx
```diff
@@ -118,7 +118,9 @@ export function PipelinesTable() {
 
   const mappedRows = useMemo(
     () =>
-      rows.map((run) => <PipelinesTableTr key={run.merged_run_id} run={run} />),
+      rows
+        .filter((run) => run.trigger)
+        .map((run) => <PipelinesTableTr key={run.merged_run_id} run={run} />),
```
Venefilyn, line 123: Checked into things a bit more, I think we can do better by displaying the rows but simply updating the `<PipelinesTableTr>`:

```tsx
<strong>
  {run.trigger && "project_url" in run.trigger ? (
    <ForgeIcon url={run.trigger.project_url} />
  ) : (
    <ForgeIconByForge />
  )}{" "}
  <Link to={`/pipeline/${run.merged_run_id}`}>
    {run.trigger && "project_url" in run.trigger ? (
      <>
        <TriggerSuffix trigger={run.trigger} />
      </>
    ) : (
      <>{run.merged_run_id}</>
    )}
  </Link>
</strong>
```

Venefilyn (changes requested):

frontend/src/components/pipeline/PipelinesTable.tsx
lbarcziova: that was my initial solution, but I believe these are really occasionally malformed data that should not be displayed at all

Venefilyn: Ack! I think this should be fixed at the service level as well to ensure this won't happen

Venefilyn (approved):
LGTM

lachmanfrantisek:
@lbarcziova ,  @Venefilyn Thanks both of you!

PR #449, opened 2024-09-03
chore: migrate prettier and eslint to biome

Venefilyn:
As Prettier pre-commit no longer works and causes issues with pnpm moving away from it would help fix two issues as there was also an issue with not being able to upgrade prettier version due to the limitation above.

There are alternatives to prettier as a substite or we can move away from pre-commit. However, moving to a different prettier alternative does make it a lot easier for us as prettier itself is very opinionated and quite slow to run, and also causes issues with eslint.

dprint was another alternative but doesn't seem as maintained or supported by more contributors. Biome can run both as a formatter and a linter, while also taking care of import order in files.

TODO:

- [ ] ‹fill in›

Supercedes
- #418 
- #419 

RELEASE NOTES BEGIN
Change dashboard to use Biome over prettier and eslint
RELEASE NOTES END


lbarcziova (approved):
looks good :+1:  thanks

PR #450, opened 2024-09-04
fix: builds failing due to pnpm

Venefilyn:
With the new restructure to use workspaces with pnpm we need to run the install within the workspace itself rather than in /frontend. We need the workspace as biome CI will work better when we have the pnpm packages in the project root

RELEASE NOTES BEGIN
Fix builds failing due to incorrect folder
RELEASE NOTES END

### DIFF
--- a/files/ansible/recipe.yaml
+++ b/files/ansible/recipe.yaml
@@ -14,7 +14,6 @@
 
     - name: install node modules
       command:
-        chdir: "{{ packit_dashboard_path }}"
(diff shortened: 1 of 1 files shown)


lbarcziova (approved):
lgtm, thanks!

I can see the [deploy-preview](https://github.com/packit/dashboard/actions/runs/10704603963/job/29677997651?pr=450) is also failing, could you have a look please?

Venefilyn:
@lbarcziova might be fixed in #451

PR #452, opened 2024-09-11
fix: builds failing due to pnpm

Venefilyn:
Follow up commit of 4048af77f5542440874fa8b50288dcad3843f373

We now make sure we're in the project root directory during installation rather than default folder. Previously we used frontend/ but that is no longer necessary as we can do it within the project root.


nforro:
Is this to fix for instance [this build error](https://github.com/packit/dashboard/actions/runs/10884283546/job/30199234917#step:5:619)? If so, can we merge this? There hasn't been a successful dashboard image build for almost 2 weeks.

Venefilyn:
Yeah it should fix it, hopefully :crossed_fingers:

PR #454, opened 2024-09-16
Updated Dockerfile with pnpm info

Venefilyn:
This adds pnpm files to the project directory within the container so that everything can be built correctly

This also removes the incorrectly set .dockerignore with lock files


nforro (approved):
LGTM. Let's see if it helps :slightly_smiling_face:

PR #463, opened 2024-10-01
Add missing states to status labels

### LINKED ISSUE #460: Use pending icon for pending states
### Description

Currently if a job is in progress, it has a sort of "i" icon:
image
This looks very similar to the error icon:
image
Both appear as bold vertical bars, with maybe a pixel gap somewhere.

As neither of these actually directly display the state directly (you need to hover to get the tooltip), and the only other differentiation is the colour, it's difficult to realize the difference between them.

It would perhaps be better to use the `pficon-pending` or `pficon-in-progress` icons, as in:
https://www.patternfly.org/components/progress-stepper/#with-custom-icons

### Benefit

Clearer differentiation between job states.

### Importance

_No response_

### Workaround

- [ ] There is an existing workaround that can be used until this feature is implemented.

### Participation

kapr200:
New colors and icons for:
* running
* pending
* skipped

Fixes #460

TODO:

- [ ] ‹fill in›

Fixes

Related to

Merge before/after

RELEASE NOTES BEGIN

Previously the pending status had same color and icon as skipped status.

Now we added the icons and colors for those two statuses, that are different from each other, and this had changed also for the running state.

RELEASE NOTES END

### DIFF
--- a/frontend/src/components/statusLabels/StatusLabel.tsx
+++ b/frontend/src/components/statusLabels/StatusLabel.tsx
@@ -2,9 +2,12 @@
 // SPDX-License-Identifier: MIT
 
 import {
+  AngleDoubleRightIcon,
   CheckCircleIcon,
   ExclamationCircleIcon,
   ExclamationTriangleIcon,
+  HourglassHalfIcon,
+  InProgressIcon,
   InfoCircleIcon,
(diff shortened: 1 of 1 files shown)


Venefilyn (approved):
image

LGTM! Thanks!

PR #465, opened 2024-10-02
chore: add Vitest

Venefilyn:
TODO:

- [ ] ‹fill in›

Fixes

Related to

Merge before/after

RELEASE NOTES BEGIN

RELEASE NOTES END

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,6 +58,9 @@ check:
 test_image: files/ansible/install-deps.yaml files/ansible/recipe-tests.yaml
 	$(CONTAINER_ENGINE) build --rm -t $(TEST_IMAGE) -f Dockerfile.tests .
 
+test_frontend:
+	cd frontend && pnpm run test
+
 check_in_container: test_image
 	$(CONTAINER_ENGINE) run --rm \
 		--security-opt label=disable \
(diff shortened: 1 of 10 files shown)


mfocko (approved):
So the lockfile got removed in #473 just to be reintroduced here…

Saving the HTML generated by React sounds like something we do in ogr for git-forge responses for tests, but at the same time I wonder what's the point of testing here, when bumping a release might change the expected result. How are you even supposed to tell that something's broken when there's a constant flow of changes from the dependencies…

Venefilyn:
@mfocko 

> So the lockfile got removed in #473 just to be reintroduced here…

There I removed a duplicate, no longer used, file as we had one in two spots. Bit confusing for sure but we should only have the one modified in this PR

> Saving the HTML generated by React sounds like something we do in ogr for git-forge responses for tests, but at the same time I wonder what's the point of testing here, when bumping a release might change the expected result. How are you even supposed to tell that something's broken when there's a constant flow of changes from the dependencies…

In most scenarios it should not change, and is an indicator that things render correctly. We can setup scenarios for how things should look like given specific inputs, such as if the API server is down, one request succeeded but the other one failed etc.

So for the snapshots that render React it is small enough that any changes should only happen when code itself is changed, and if it is changed on each dependency update, that should be addressed for sure so it doesn't happen again (if nothing visual is changing that is). But PatternFly tends to keep visual changes to a minimum.

We can also do image/pixel tests if we want to instead, but that takes up a bit more storage

mfocko:
> There I removed a duplicate, no longer used, file as we had one in two spots. Bit confusing for sure but we should only have the one modified in this PR

OH! that's not a new file, that's 1k new lines in the existing one /o\

PR #471, opened 2024-10-11
fix(Copr): Add chroot to copr enable

### LINKED ISSUE #470: misleading instructions how to use the copr builds
### What happened? What is the problem?

Ohai,

when you have packit build a PR, the users get a link to the dashboard `https://dashboard.packit.dev/jobs/copr/:id` (e.g. https://dashboard.packit.dev/jobs/copr/1933132) that contains a message how to use the resulting builds, however, those are (possibly) misleading for users not very well versed in copr/dnf.

## `dnf copr enable {data.copr_owner}/{data.copr_project}` without `chroot`

Right now, the instructions say to call `sudo dnf copr enable packit/theforeman-foreman_maintain-936`, which on EL8 will try to use the `epel-8-x86_64` chroot by default (and `epel-9-x86_64` on EL9), but Packit didn't build for those chroots (the linked build is for `rhel-8-x86_64`) and the user ends up with an error:
```
Error: It wasn't possible to enable this project.
Repository 'epel-8-x86_64' does not exist in project 'packit/theforeman-foreman_maintain-936'.
Available repositories: 'rhel-9-x86_64', 'rhel-8-x86_64'
```

Given Packit knows which chroot the build was for, it could add that to the `dnf copr` invocation, right?

## `dnf install -y <all packages>`

This might be a tad specific to @theforeman packages, but we do have quite a lot sub-packages that get built, but are not needed on a user system (they contain code and dependencies that are needed to build plugins etc).

The instructions on the copr job view tell the user (who might be unaware of our package split!) to `dnf install -y all-the-packages-that-were-built` which in the best case results in the user installing things they don't need and in the worst case results in dependency errors from `dnf` as we don't ship all packages required for *building* in our main repos.

It would be cool if the wording would point out that this might not what you need, or allow projects to override this messaging in their `packit.yml`

### What did you expect to happen?

_No response_

### Example URL(s)

https://dashboard.packit.dev/jobs/copr/1933132

### Steps to reproduce

```bash
1. tell users "take the packit build from <PR link>"
2. see them following the instructions literally
```

### Workaround

- [X] There is an existing workaround that can be used until this issue is fixed.

### Participation

Venefilyn:
This improves the installation instructions to align better with
installing the specific build we're displaying. If not for this you
couldn't install rhel9 specific chroots as dnf copr would assume epel9

image showing chroot added to dnf copr enable

Fixes #470

RELEASE NOTES BEGIN
Add chroot to Copr build installation instructions
RELEASE NOTES END

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -62,5 +62,6 @@
   "engines": {
     "node": ">=20",
     "pnpm": ">=9"
-  }
+  },
+  "packageManager": "pnpm@9.12.1+sha512.e5a7e52a4183a02d5931057f7a0dbff9d5e9ce3161e33fa68ae392125b79282a8a8a470a51dfc8a0ed86221442eb2fb57019b0990ed24fab519bf0e1bc5ccfc4"
 }
--- a/frontend/src/components/copr/CoprBuild.tsx
+++ b/frontend/src/components/copr/CoprBuild.tsx
@@ -92,7 +92,8 @@ export const CoprBuild = () => {
                   </ListItem>
                   <ListItem>
                     <code>
-                      sudo dnf copr enable {data.copr_owner}/{data.copr_project}
+                      sudo dnf copr enable {data.copr_owner}/{data.copr_project}{" "}
+                      {data.chroot}
                     </code>
                   </ListItem>
                   <ListItem>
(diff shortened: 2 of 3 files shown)


lbarcziova (approved):
lgtm, thanks!

PR #481, opened 2024-11-06
feat: More build info in Testing Farm runs

Venefilyn:
This adds more information while also taking away some extra steps
relating to the Copr builds associated with the testing farm results.

Now looks like this
image

RELEASE NOTES BEGIN
Improve UX of Copr builds in Testing Farm runs
RELEASE NOTES END


lbarcziova (approved):
much better, thanks!

PR #485, opened 2024-11-11
Improve OpenScanHub URLs & ensured TypeScript compliance

Venefilyn:
There were some stuff missing regarding styling of the API and what went
on within components. In particular a lot of additions that were not
changed after copying from another component.

As for the API interfaces they were missing the submitted_time which
showed errors within the components themselves.

Also URLs and titles have been renamed to indicate OpenScanHub more
explicitly as we do not have to use abbreviations.

RELEASE NOTES BEGIN
Rename OpenScanHub URLs and improved styling
RELEASE NOTES END


lbarcziova (approved):
thanks for the improvements!

PR #487, opened 2024-11-13
feat(Copr): Improved installation instructions

### LINKED ISSUE #470: misleading instructions how to use the copr builds

Venefilyn:
This splits the packages up in easier chunks for installation.

Fixes #470

image

RELEASE NOTES BEGIN
Improve Copr installation instructions
RELEASE NOTES END

### DIFF
--- a/frontend/src/components/copr/CoprBuild.tsx
+++ b/frontend/src/components/copr/CoprBuild.tsx
@@ -4,7 +4,13 @@
 import {
   Card,
   CardBody,
+  ClipboardCopy,
   Content,
+  DataList,
+  DataListCell,
+  DataListItem,
+  DataListItemCells,
+  DataListItemRow,
   List,
   ListItem,
   PageSection,
@@ -12,6 +18,7 @@ import {
 } from "@patternfly/react-core";
 
 import { useQuery } from "@tanstack/react-query";
+import { useEffect, useState } from "react";
 import { CoprBuildPackage } from "../../apiDefinitions";
 import { coprBuildQueryOptions } from "../../queries/copr/coprBuildQuery";
 import { Route as CoprRoute } from "../../routes/jobs_/copr.$id";
@@ -25,6 +32,12 @@ export const CoprBuild = () => {
   const { id } = CoprRoute.useParams();
 
   const { data, isError, isLoading } = useQuery(coprBuildQueryOptions({ id }));
+
+  const [packagesToInstall, setPackagesToInstall] = useState<string[]>([]);
+
+  useEffect(() => {
+    if (data) setPackagesToInstall(getPackagesToInstall(data.built_packages));
+  }, [data?.built_packages]);
 
   // If backend API is down
   if (isError) {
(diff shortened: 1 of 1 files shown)


lachmanfrantisek (approved):
Nice, this is handy!

frontend/src/components/copr/CoprBuild.tsx
```diff
@@ -79,32 +92,77 @@ export const CoprBuild = () => {
...
+                  <Title headingLevel="h3">Repository</Title>
+                  <List isPlain>
+                    <ListItem>
+                      <ClipboardCopy
+                        hoverTip="Copy"
+                        clickTip="Copied"
+                        variant="inline-compact"
+                        isCode
+                      >
+                        sudo dnf install -y dnf-plugins-core
```
nforro, line 109: Just a note, on F41+ the plugin is in `dnf5-plugins`, but it should be installed automatically.

lbarcziova (approved):
looks good!

do you know what might be causing the deploy-preview to fail? (briefly checked and it happened for past few PRs already)

frontend/src/components/copr/CoprBuild.tsx
nforro: `dnf -y install 'dnf-command(copr)'` would be safer, though funnily enough `dnf5-plugins` doesn't provide it (but I'm sure they'll fix it eventually).

Venefilyn:
@lbarcziova should be fixed in #488

frontend/src/components/copr/CoprBuild.tsx
nforro: OTOH it seems to work everywhere (where there is `dnf`). @packit/the-packit-team WDYT?

```suggestion
                        sudo dnf install -y 'dnf*-command(copr)'
```

lbarcziova: I like the suggestion

Venefilyn: @nforro @lbarcziova added!

PR #488, opened 2024-11-13
chore: Update pnpm cache location

Venefilyn:
With the removal of an unused pnpm-lock with workspaces the workflow started to fail as it doesn't use the newer file location.

RELEASE NOTES BEGIN
Fixed PR preview yet again
RELEASE NOTES END


Venefilyn:
Workflow will fail until merged

lbarcziova (approved):
thanks!

PR #500, opened 2024-12-04
Change pagination from infinite loading to paged pagination

Venefilyn:
This now has a context that each component under the provider can use.
With this we reduce the code we need to write to handle state handling
between components.

Infinite pagination has been replaced with per-page pagination which
should suit people better than infinitely loading.

Before
image

After
image

TODO:
- [ ] Decide if we want this pagination
- [ ] Migrate other components
- [ ] Remove the end of pagination handling due to bug (insert issue link here)

RELEASE NOTES BEGIN
Change pagination from infinite loading to paged pagination.
RELEASE NOTES END

### DIFF
--- a/frontend/src/components/bodhi/BodhiUpdatesTable.tsx
+++ b/frontend/src/components/bodhi/BodhiUpdatesTable.tsx
@@ -1,7 +1,7 @@
 // Copyright Contributors to the Packit project.
 // SPDX-License-Identifier: MIT
 
-import { useMemo } from "react";
+import { useState } from "react";
 
 import {
   Table,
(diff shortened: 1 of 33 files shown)


lbarcziova:
packit/packit-service#2505 might affect this.

lbarcziova:
Once packit/packit-service#2645 is implemented, this should be changed adjusting to that, right?

Venefilyn:
> Once [packit/packit-service#2645](https://github.com/packit/packit-service/issues/2645) is implemented, this should be changed adjusting to that, right?

@lbarcziova Might've missed this if I didn't answer in a meeting. But pagination in general would change so this PR doesn't introduce much extra code that would need to be changed

lbarcziova (approved):
looks great, thanks a lot! 🚀 

I have played with it a bit and didn't spot any issues, let's get it to staging and see during the week

PR #502, opened 2024-12-18
Fix OpenScanHub schema not reflecting API data

Venefilyn:
There was a discrepancy between what the API should return and what is
being returned. This fixes some of the fields that can be null and
handles the case where that occurs.

Related to https://dashboard.packit.dev/jobs/openscanhub/3694

RELEASE NOTES BEGIN
Update Packit's OpenScanHub API endpoint schema
RELEASE NOTES END


majamassarini (approved):
LGTM thanks!

lbarcziova (approved):
thanks!

PR #524, opened 2026-02-06
Cleaning up tests and dead code

jpodivin:
These things slightly bothered me when I worked on my previous PR. I could have performed steeper cuts to this, but parts may still be of some use in the future.

Changes: 

- Fixed path to `package.json` in test docker file
- Removed attempt to copy non-existant templates in dockerfile
- Removed `utils` module, since the only function wasn't used anymore
- Removed the `test_architecture_image` test case, as there is no architecture diagram
- Removed commented out and dead test_requre files

I have kept data files used for testing and fixtures. Although I suspect that no tests are going to ever use them.


mfocko (approved):
LGTM, could you please rebase?

PR #532, opened 2026-05-20
Pin GitHub Actions to commit hashes for security

majamassarini:
Pin all actions to specific commit SHAs to prevent supply chain attacks and ensure reproducible builds.

Assisted-By: Claude Sonnet 4.5 <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -17,7 +17,7 @@ jobs:
             image: dashboard
 
     steps:
-      - uses: actions/checkout@v3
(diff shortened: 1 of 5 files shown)


Venefilyn (approved):
Can probably update dependabot to create PRs when there are new stuff

majamassarini:
> Can probably update dependabot to create PRs when there are new stuff

Thanks for the review! And yes it would be nice to have it, but on the other hand, we don't have a way to test wether the new versions break the actions and this is why we are using really old versions and I haven't even update them.

PR #534, opened 2026-07-02
Add pyproject.toml for Python dependency management

### LINKED ISSUE #201: Flask server has no dependency handling
While setting up the server for local development (without container) I noticed that there is nothing to indicate what you need to install when it comes to dependencies or what version you need. Ideally we should be using something similar to yarn or npm to handle dependencies and lock things down, even if it is minor

I propose that we should use poetry to set this up

```py
⬢ [fedora-toolbox:37] ❯ make run-dev-flask 
FLASK_ENV=development FLASK_APP=packit_dashboard.app flask-3 run --host=0.0.0.0
Usage: flask-3 run [OPTIONS]
Try 'flask-3 run --help' for help.

Error: While importing 'packit_dashboard.app', an ImportError was raised:

Traceback (most recent call last):
  File "/usr/lib/python3.11/site-packages/flask/cli.py", line 218, in locate_app
    __import__(module_name)
  File "/var/home/spytec/projects/dashboard/packit_dashboard/app.py", line 3, in <module>
    from packit_dashboard.api.routes import api
  File "/var/home/spytec/projects/dashboard/packit_dashboard/api/routes.py", line 2, in <module>
    from packit_dashboard.utils import return_json
  File "/var/home/spytec/projects/dashboard/packit_dashboard/utils.py", line 1, in <module>
    import requests
ModuleNotFoundError: No module named 'requests'
```

majamassarini:
Resolves #201 by adding proper dependency handling for local development using `pyproject.toml` (following the same pattern as the packit package).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,6 +25,9 @@ classifiers = [
     "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
     "Topic :: Software Development",
 ]
 keywords = [
@@ -35,7 +38,6 @@ dependencies = [
     "Flask >= 2.0.0",
     "flask-cors >= 3.0.0",
     "flask-talisman >= 1.0.0",
-    "cachetools >= 4.0.0",
     "requests >= 2.20.0",
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,53 @@
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[project]
+name = "packit-dashboard"
+# Using static version for now since the repo doesn't have version tags yet.
+# TODO: Switch to dynamic versioning with hatch-vcs when version tags are added.
+version = "0.1.0"
+authors = [
+    { name = "Red Hat", email = "hello@packit.dev" },
+]
+description = "Dashboard for Packit Service"
+readme = "README.md"
+license = "MIT"
+license-files = ["LICENSE"]
+requires-python = ">=3.9"
+classifiers = [
+    "Intended Audience :: Developers",
+    "License :: OSI Approved :: MIT License",
+    "Operating System :: POSIX :: Linux",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3 :: Only",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Topic :: Software Development",
+]
+keywords = [
+    "packit",
+    "dashboard",
+]
+dependencies = [
+    "Flask >= 2.0.0",
+    "flask-cors >= 3.0.0",
+    "flask-talisman >= 1.0.0",
+    "cachetools >= 4.0.0",
+    "requests >= 2.20.0",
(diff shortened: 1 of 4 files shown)


nforro (commented):
Here are some points from Claude review that I think are valid:

  ### 1. CONTRIBUTING.md wording could be misleading

  The PR changes the CONTRIBUTING.md from:

  ▎ make install-dependencies automatically installs python dependencies and pnpm.

  To:

  ▎ Python dependencies are managed using pyproject.toml.

  This is slightly misleading — make install-dependencies still installs RPM packages, not from pyproject.toml. The
  pyproject.toml is only used in the pip/virtualenv path. The new wording implies pyproject.toml is the primary/sole
  mechanism, which isn't true for the RPM-based workflow.

  Suggestion: Something like:

  ▎ Python dependencies are declared in pyproject.toml and can be installed either via system packages or pip.

  ### 2. No version pins on dependencies

  All dependencies are unpinned ("Flask", "requests", etc.). The original issue #201 specifically called out wanting to "lock
  things down" version-wise. While the packit project also uses mostly unpinned deps, it does pin some (e.g., "marshmallow  >= 3.18.0", "bodhi-client >= 7.0.0").

  This isn't necessarily wrong for a dashboard app with few dependencies, but it's worth acknowledging that this partially
  addresses #201 — it defines what to install but doesn't lock versions. A follow-up with a lock file or minimum version
  bounds would fully address the spirit of the issue.

  ### 3. version = "0.1.0" is hardcoded

  The packit project uses dynamic = ["version"] with hatch-vcs to derive the version from git tags. This PR hardcodes
  "0.1.0". If the dashboard doesn't publish to PyPI, this is fine. But if there's any plan to version it properly in the
  future, it would be better to set this up with hatch-vcs now (or at least leave a comment about the intent).

pyproject.toml
```diff
@@ -0,0 +1,53 @@
...
+]
+keywords = [
+    "packit",
+    "dashboard",
+]
+dependencies = [
+    "Flask >= 2.0.0",
+    "flask-cors >= 3.0.0",
+    "flask-talisman >= 1.0.0",
+    "cachetools >= 4.0.0",
```
nforro: Is this really used somewhere?

pyproject.toml
```diff
@@ -0,0 +1,53 @@
...
+# Using static version for now since the repo doesn't have version tags yet.
+# TODO: Switch to dynamic versioning with hatch-vcs when version tags are added.
+version = "0.1.0"
+authors = [
+    { name = "Red Hat", email = "hello@packit.dev" },
+]
+description = "Dashboard for Packit Service"
+readme = "README.md"
+license = "MIT"
+license-files = { paths = ["LICENSE"] }
```
nforro: This syntax has been deprecated.

```suggestion
license-files = ["LICENSE"]
```

pyproject.toml
```diff
@@ -0,0 +1,53 @@
...
+classifiers = [
+    "Intended Audience :: Developers",
+    "License :: OSI Approved :: MIT License",
+    "Operating System :: POSIX :: Linux",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3 :: Only",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
```
nforro, line 27: Are 3.12, 3.13 and 3.14 not supported?

nforro (approved):
It seems there is a mismatch between what's installed by ansible and `make install-dependencies` - the latter is mising python3-requests. But that could be considered to be out of scope of this PR.

PR #537, opened 2026-07-09
Display error msg for failed Log Detective runs

Jany26:
The error is displayed in the same way as the explanation:
image

I also conditioned the display on `status`.

This should be fine to merge, even though https://github.com/packit/packit-service/pull/3151 is not yet in prod. I checked and display works fine even if error_msg field is not present, or null.

RELEASE NOTES BEGIN

Display error messages for failed Log Detective runs.

RELEASE NOTES END


mfocko (approved):
lgtm

jpodivin (approved):

mfocko:
oh right, this is the project where we started experimenting with merge queues, but Zuul wasn't nuked…

PR #544, opened 2026-09-22
Display Log Detective jobs/runs triggered by the same action on the same line

Jany26:
The default `[Log Detective]` display on the dashboard will now be utilizing the new `api/log-detective/groups/` defined in https://github.com/packit/packit-service/pull/3191. 
I also moved some info to the detailed view, see images below.

EDIT: Updated the images.

Tested locally with a mock api serving the modified data.

---

The first image triggers `GET .../api/log-detective/groups?page=1&per_page=10`:
image

---

The second image triggers `GET .../api/log-detective/{ID}`:
image

Merge after https://github.com/packit/packit-service/pull/3191 is in prod.

RELEASE NOTES BEGIN

Log Detective runs are now grouped by trigger on the overview page.

RELEASE NOTES END


jpodivin (approved):

mfocko (approved):

frontend/src/apiDefinitions.ts
```diff
@@ -496,36 +496,19 @@ export interface OSHScan {
...
-  anitya_package: string | null;
 }
 
-// /api/log-detective/groups/$id
 export interface LogDetectiveGroup {
   packit_id: number;
   submitted_time: number | null;
   run_ids: number[];
-  log_detective_target_ids: number[];
+  log_detective_targets: LogDetectiveTarget[];
```
betulependule: Unless I'm mistaken, I think that you also need to replace `log_detective_target_ids` with `log_detective_targets` in `LogDetectiveGroup.tsx`. VS Code is complaining that `log_detective_target_ids` does not exist on type `LogDetectiveGroup` in that file.

frontend/src/components/logdetective/LogDetectiveResultsTable.tsx
```diff
@@ -90,36 +86,35 @@ export const LogDetectiveResultsTable = () => {
...
+                        status={target.status}
+                        target={target.target_arch}
+                        link={`/jobs/log-detective/${target.id}`}
+                      />
+                    </span>
+                  ))}
                 </Td>
                 <Td dataLabel={columnNames.commitSha}>
-                  {log_detective_result.commit_sha}
+                  {group.commit_sha?.slice(0, 7)}
```
betulependule: I'm wondering whether we could use the `SHACopy` component here. This just displays the commit SHA, but using `SHACopy` would create a clickable button next to it to enable users to copy it to clipboard very easily.

It is used for example here (both the trigger and commit SHA are below the "Koji Build Results" title)
https://dashboard.packit.dev/jobs/koji/46990

It might be nice to do this similarly for the "Log Detective Results" (to display both the trigger and commit SHA below the title), but I'm fine with keeping it the way it is if that's the visual look that what you prefer.

Jany26: agreed

frontend/src/apiDefinitions.ts
Jany26: yeah forgot to fix this, but I feel like the group view is now redundant anyway, so I'll remove it completely. thx

Jany26:
@jpodivin couple of new changes based on some nice ideas from review:
- LD job detail view now uses commit sha link and trigger as a sub-heading, similar to other dashboard modules (see updated images in PR desc)
- removed /log-detective/group/{id} view since it is basically redundant now with the changes, I don't know if anyone uses the view for anything, see below:

image

PR #545, opened 2026-09-23
fix: allow running without HTTPS

mfocko:
As part of PACKIT-5356 we need to be able to run the container without providing the SSL/TLS itself, as it will be provided by the OpenShift.

This should allow us to conditionally let the ‹mod_wsgi-express› handle the SSL/TLS, if there are certificates provided, otherwise just use plain HTTP and let the reverse proxy / OpenShift handle it.

Fixes PACKIT-5356

Assisted-by: gpt-5.6-luna medium


betulependule (approved):
LGTM 🙏

